PR #11, opened 2023-01-30
Code reformatting of MetaInsight

nabury:
Major reformatting of server.R to add more reactive functions and improve output code.

No changes to functionality.

New files added to contain functions - bugsnet_sumtb.R, plot.R and util.R
New functions added and edits to functions in fn_analysis.R


BalancedmonkeY (changes requested):
All looks good, and even covers some small things that I found annoying but never had the time to sort out - thanks!

Only request is to check the downloadbuttons.R file as I think there may need to be some changes there as it uses a lot of the functions that have been moved/altered away from the server/fn_analysis.

nabury:
downloadbuttons.R file edited to match new function calls.

Manually checked download buttons and file upload.

Nodesplit model code checked and duplicated/old code removed.

BalancedmonkeY (approved):

PR #20, opened 2023-03-22
Allow Data Changes

### LINKED ISSUE #10: Remove user uploaded data if change in continuous or binary selection on home tab
At the moment, if a user uploads their own data set and then changes between binary or continuous on the 'home' tab, the user data remains selected. Suggest changing this so that the user data is removed and the correct example data is selected when the user changes between continuous and binary on the home tab.

RyanJField:
Fixes #10 
Adds a delete data button to allow users to upload a new dataset
Adds a reload labels button to allow users to reload the appropriate default labels
Needs review and further testing prior to merge
I have done my best to not change any of the core functionality.


nabury:
I've manually tested all combinations of this that I can think of and it looks good! Suggest waiting until issue #21 is fixed before merging.

PR #23, opened 2023-04-03
Radial SUCRA code compliant with {ggplot2}

BalancedmonkeY:
The code for the Radial SUCRA plot had to be amended to be compliant with newer versions of {ggplot2}


nabury (commented):
This works on my machine - I am getting the following warning:

Warning in left_join(pairs2, studylabs, by = "mtchvar") :
  Each row in `x` is expected to match at most 1 row in `y`.
ℹ Row 1 of `x` matches multiple rows.
ℹ If multiple matches are expected, set `multiple = "all"` to silence this warning.

PR #29, opened 2023-05-30
Feature user guide for offline users

nabury:
Links from section 3d to the user guide on the wiki.

Includes a file packages.R that lists all the packages currently used in MetaInsight (if people need to install them on their machine)


RyanJField (commented):
This looks good, I successfully followed the wiki and got the app to run locally on both Windows and Mac.

A comment on the wiki: Could we put the code that users run in a code block e.g.
```
Code goes here
```

To make copy an pasting easier? Code blocks start and end with \`\`\` three back ticks

I think we might need a discussion on package management, as we could turn the app into an R package to make installing dependencies easier,  currently I'm using packrat but this would be harder to use by end-users if they were using the app locally.

Other than that this looks good.

PR #41, opened 2023-06-26
Funding

nabury:
Edited funding statement at the bottom of the home page.

- BRC funding statement added
- New CRSU logo (with new NIHR logo) added and old CRSU logo removed
- wellPanel layout changed so that the logo fills 100% of the width of the page and text sits below it (rather than side-by-side as before)
- Old CRSU logo deleted from www folder

Addresses issues #30 and #31


JanionNevill (approved):
Looks good

Issue #34, opened 2023-06-14
Add SEO meta tags


nabury:
Add SEO meta tags so the current version of MetaInsight is higher on search engines

PR #43, opened 2023-06-30
Add Meta Tags, Favicon, OpenGraph Move CSS #34

RyanJField:
I have added the SEO Tags per (#34)

I have also moved the CSS to a separate file, and added a favicon.


ui.R
```diff
@@ -1,794 +1,797 @@
...
+  dashboardBody(
+  tags$head(
+    # load custom stylesheet
+    # To ensure white background and no horizontal scroll bars on ranking panel
+    tags$link(rel = "stylesheet", type = "text/css", href = "app.css"),
+    includeHTML("www/favicon/favicon.html"),
+    tags$meta(name="description", content="A interactive web tool for network meta-analysis (NMA) that leverages established analysis routines"),
+    tags$meta(name="keywords", content="MetaInsight, NMA, Network Meta Analysis, App"),
+    tags$meta(property="og:title", content="Meta Insight: V4.1.0"),
+    tags$meta(property="og:description", content="A interactive web tool for network meta-analysis (NMA) that leverages established analysis routines"),
```
JanionNevill, line 45: Should be "***An*** Interactive web tool..."

ui.R
```diff
@@ -1,794 +1,797 @@
...
+  dashboardHeader(disable = TRUE),
+  dashboardSidebar(disable = TRUE),
+  dashboardBody(
+  tags$head(
+    # load custom stylesheet
+    # To ensure white background and no horizontal scroll bars on ranking panel
+    tags$link(rel = "stylesheet", type = "text/css", href = "app.css"),
+    includeHTML("www/favicon/favicon.html"),
+    tags$meta(name="description", content="A interactive web tool for network meta-analysis (NMA) that leverages established analysis routines"),
+    tags$meta(name="keywords", content="MetaInsight, NMA, Network Meta Analysis, App"),
```
JanionNevill, line 43: do we want/need "meta analysis" as a separate keyword, or does it already match due to it being a subset?

www/app.css
```diff
@@ -0,0 +1,6 @@
+.content-wrapper {
+  background-color: #fff;
```
JanionNevill, line 2: Is that a standard way of writing a hex colour? Should it be `#ffffff`?

JanionNevill (changes requested):

www/app.css
RyanJField: I didn't write this but it is an accepted way of writing white. See [short notation](https://www.w3.org/TR/css-color-3/#numerical) in the W3C recommendations.

ui.R
RyanJField: They should probably be split, but most search engines ignore them anyway, but they are useful to know what keywords you should be targeting.

RyanJField: I have split these now.

ui.R
RyanJField: Thank you, changed

JanionNevill (approved):

nabury (approved):

Issue #40, opened 2023-06-23
Remove full update history tab


nabury:
Before next merge of dev branch to main branch:

- remove the full update history tab
- delete associated files in www folder
- change 'Click here to view a full update history of MetaInsight' on home page to point to MetaInsight wiki
- add all future updates to wiki only 

Reasoning: wiki is easier & faster to update than using pdf files

PR #45, opened 2023-07-04
Update history

nabury:
Removed the full update history tab as in issue #40 

- remove the full update history tab
- delete associated files in www folder
- change 'Click here to view a full update history of MetaInsight' on home page to point to MetaInsight wiki

Also removed repeated links to GitHub from the home page


JanionNevill (approved):
Looks good

PR #48, opened 2023-07-17
MI-2 summary forest plots

JanionNevill:
Added summary forest plots into frequentist analysis tab. Also tidied up a few quick wins.


R/SummaryForestPlotPane.R
nabury, line 1: Could you ask Ryan to also review the modules? (the code looks good and is definitely readable but I don't have any experience of writing modules myself so he may pick up any issues I can't)

R/SummaryForestPlotsTab.R
```diff
@@ -0,0 +1,51 @@
+
```
nabury, line 1: The plots are quite small when viewed side by side on a laptop screen - I would like to get some other people's thoughts on having the plots side-by-side or top-to-bottom.

R/summary_forest_plot.R
nabury, line 1: Have you edited any of this file?

JanionNevill: I made small changes to the code to get it to work for a generalised data set. Nothing major though, it was mostly copied out of `plot_functionsRKO.R` or taken from the code shared by you and Alex.

R/SummaryForestPlotsTab.R
JanionNevill: That's a good point about the size. Everything else in the app is side-by side, so that becomes part of the user's expectations using the app. I don't have a strong suggestion on how to lay them out better without breaking that user experience.

nabury: Some of the tabs are top-to-bottom i.e. tab 2b which is not that dissimilar to tab 2d.

R/summary_forest_plot.R
nabury: Are there any particular bits of new code you want me to look at? (GitHub is displaying the whole file as new)

JanionNevill: I've just looked over it, and there's nothing specific that I've really changed. Maybe the best option is to try to stress-test it in the app by playing with inputs to see if it breaks?

R/SummaryForestPlotsTab.R
RyanJField: I think we need to discuss making plots reactive (CSS) so they grow an shrink with the window. Another idea I think would be good is to have an on click event on the plots that shows that plot as a modal (in window pop up), and adding an on:hover CSS to change the curser to a magnifying glass.

R/SummaryForestPlotPane.R
RyanJField: > Could you ask Ryan to also review the modules? (the code looks good and is definitely readable but I don't have any experience of writing modules myself so he may pick up any issues I can't)

The modules look good to me.

R/SummaryForestPlotsTab.R
nabury: I don't disagree @RyanJField , but I think that should be a separate issue (to be implemented across the app).
Having thought about the layout more, I would argue that the summary forest plots should be laid out top-to-bottom because the width of the plot increases with increasing numbers of treatments.

JanionNevill: I've made them atop one another, rather than side-by-side. I like the idea of the better plot interactions, but I agree with Naomi that it would be best as a separate task (Which I think is what you meant).

nabury (approved):

PR #50, opened 2023-07-27
Mi 4 remove bayesian analysis

nabury:


JanionNevill:
I've not done a code review yet, but what I'm finding is that when I first load the app, the Bayesian forest plot doesn't appear when the model is run.
#### Steps to Reproduce:
- Launch app
- Go straight to Bayesian tab in data analysis page
- Click the button to run the all-studies model

After this, the plot download options flash up, then immediately disappear. Running the model a second time shows the plot and associated UI. The same seems to happen to the filtered model.

JanionNevill:
I've also found that unticking a filtered study does not clear the filtered analysis, but ticking a study does. Can you investigate this please?

server.R
```diff
@@ -719,6 +720,266 @@ shinyServer(function(input, output, session) {
     bayesian_model(sub = TRUE, data(), treatment_list(), input$metaoutcome, input$exclusionbox, 
                    outcome_measure(), input$modelranfix, reference_alter())
   })
+  
+  # Hide the outdated Bayesian visualisations when the data is changed 
+  # or the inputs to the outcome, treatment ranking or model radio buttons are changed
+  observeEvent(c(data(), input$outcomeCont, input$outcomebina, input$rankopts, input$modelranfix), {
```
JanionNevill, line 726: I'm thinking forward to when the app is modularised.
Can you change these blocks such that there are 2 `reactiveVal`s, named something like *model_cleared* and *sub_model_cleared*. These reactives would be set to `TRUE` initially, and when the various inputs change, and would be set to `FALSE` when the button is clicked to run the model.
The `observeEvent` blocks to hide the content would use those reactives as triggers.

When it comes to modularising the code, the 2 reactives wil be passed through to the various modules for each page, and each will be responsible only for hiding its own content.

server.R
```diff
@@ -719,6 +720,266 @@ shinyServer(function(input, output, session) {
...
+    # Elements in tab 3a left column
```
JanionNevill, line 727: There's a lot of bloat added to the code by adding all of these IDs. Can you factor them out into a separate file which assigns lists of IDs to be hidden by all changes, and to be hidden by filters, etc.
The code is `server.R` would then just use those lists within these blocks, either iterating over them, or preferably passing the whole list into the show/hide function, if they support that.

JanionNevill (changes requested):

PR #54, opened 2023-08-07
Added data upgrade panel

JanionNevill:
This can be extended if we do any more data changes before the next release.

### DIFF
--- a/R/DataInputPanel.R
+++ b/R/DataInputPanel.R
@@ -23,7 +23,12 @@ data_input_panel_ui <- function(id) {
       )
     ),
     div(class = "clearfix"),
-    selectizeInput(inputId = ns('reference_treatment'), label = 'Select Reference Treatment', choices = c())
+    selectizeInput(
+      inputId = ns('reference_treatment'),
+      label = 'Select Reference Treatment',
+      choices = c()
+    ),
+    data_upgrade_panel_ui(id = ns("data_upgrade_panel"))
   )
 }
 
@@ -39,6 +44,8 @@ data_input_panel_ui <- function(id) {
 data_input_panel_server <- function(id, metaoutcome, continuous_file = 'Cont_long.csv', binary_file = 'Binary_long.csv') {
   moduleServer(id, function(input, output, session) {
     ns <- session$ns
+    
+    data_upgrade_panel_server(id = "data_upgrade_panel")
     
     # Create a definable reactive value to allow reloading of data
     reload <- reactiveVal(FALSE)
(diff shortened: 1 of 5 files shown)


R/DataUpgradePanel.R
```diff
@@ -0,0 +1,100 @@
...
+      )
+    })
+    
+    treatments_df <- reactive({
+      return(CreateTreatmentsDataFrame(input$treatment_names))
+    })
+    
+    treatments_df <- reactive({
+      return(CreateTreatmentsDataFrame(input$treatment_names))
+    })
```
BalancedmonkeY, line 60: Duplicated code?

tests/testthat/test-DataUpgrade.R
```diff
@@ -0,0 +1,165 @@
...
+  expect_equal(colnames(treatment_df), c("Number", "Label"),
+               label = format_vector_to_string(colnames(treatment_df)))
+  expect_equal(nrow(treatment_df), 5,
+               label = nrow(treatment_df))
+  expect_equal(treatment_df$Number, 1:5,
+               label = format_vector_to_string(treatment_df$Number))
+  expect_equal(treatment_df$Label, c("Egg", "Flour", "Sugar", "Butter", "Cinnamon"),
+               label = format_vector_to_string(treatment_df$Label))
+})
+
```
BalancedmonkeY, line 46: Should there be another test for when there are spaces in the middle of an item, e.g. 'Baking Powder', as a treatment name like this would cause errors?

R/DataInputPanel.R
BalancedmonkeY, line 1: Line 10 - maybe add something mentioning that there is the data upgrader option to help.

BalancedmonkeY (changes requested):

R/DataUpgradePanel.R
JanionNevill: Removed. No idea how that snuck in!

tests/testthat/test-DataUpgrade.R
JanionNevill: I've added handling for this

R/DataUpgrade.R
```diff
@@ -0,0 +1,55 @@
...
+      }
+      upgraded[[col]] <- treatment_df$Label[match(upgraded[[col]], treatment_df$Number)]
+    }
+  }
+  return(upgraded)
+}
+
+#' Create data frame containing treatment IDs and names.
+#' 
+#' @param treatment_names_string String containing treatment names, separated by commas
```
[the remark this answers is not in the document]
JanionNevill: I intentionally made this a simple list because the whole columns thing was the pain point. This will require a small amount of reformatting for people, but it'll be much more robust, so I think it's worth the effort. It's very difficult to mess up a comma separated list, and I've done some data tidying anyway for any minor errors.

R/DataInputPanel.R
JanionNevill: Done

BalancedmonkeY (approved):

PR #62, opened 2023-08-16
Hotfix download buttons

nabury:


tommorris168 (changes requested):
The long dataset now download properly.
The wide datasets are the same as the long, i.e. not wide.

Also, could the datasets be named to indicate what's in them, because they are all called "MetaInsightdataLONG", and the user has to rename them if they download more than one.

PR #65, opened 2023-08-25
Removed unused plotting code

JanionNevill:


BalancedmonkeY (approved):
All file removals and modifications make sense. I've checked the functions in PlotFunctionsRKO.R and do not think any of them are being used in the app anymore.

PR #83, opened 2023-09-25
MI-18 reinstall sfp

JanionNevill:
Reinstated summary forest matrix plot.


R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+
+
+#' Draw the elements of the summary forest matrix plot.
+#'
+#' @param stytitle Title of plot
+#' @param ntx Number of treatments
+#' @param lstx Vector of treatment names
+#' @param mtc Meta-analysis data for direct and indirect evidence
+#' @param ma Meta-analysis data for direct evidence only
+#' @param bpredd TRUE if confidence interval to be plotted as error bars
```
BalancedmonkeY, line 240: I think this should be 'predictive interval' (based on name)
Same on line 579

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+    }
+  }
+  
+  #Identify original number of tx in analysis - for use in rankjar function
+  ori.ntx <- length(mtc$rkgram[, 1]) / ntx
+  
+  #Draw graphs along the diagonal
+  rankogram(ntx, mtc$rkgram[, 1])
+  
+  #Insert the graph title and x-axis title
```
BalancedmonkeY, line 357: MetaInsight also does RR Risk Ratio, RD Risk Difference, and SMD Standardised mean difference. Do you know of any specific reason why these outcomes aren't included (which can just be that you found documentation that the plot is limited to OR, HR, and MD)?

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+#' @param mtorg Sorting order matrix
+#'
+#' @return Updated meta-analysis data
+ma.sortres <- function(ma, mtc, mtorg) {
+  #Re-calculate estimates after inverting the reference group, using mtorg[, 4 = inv]
+  tmp.lor <- mtorg[, 4] * (ma$lor[, 4:7])
+  tmp.or <- mtorg[, 4] * (ma$or[, 4:7])
+  tmp.predint <- mtorg[, 4] * (ma$predint[, 4:7])
+  
+  #swap 25% & 75% estimates for those that we inverted the reference group
```
BalancedmonkeY, line 450: 2.5% and 97.5%?
Same on line 495

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+  
+  if (!bkey) {
+    plt.adj <- 0
+  } else if (p.only < ntx) {
+    plt.adj <- 2
+  } else {
+    plt.adj <- 1
+  }
+  
+  sp.order <- "Interventions are displayed sorted by median rank."
```
BalancedmonkeY, line 598: Is this true - or is it being sorted by SUCRA?

I can see in the code that the ranks presented are from `mtc$rank[i %% ntx, 3]`, what value is this? 

Comment also relates to line 630

BalancedmonkeY (changes requested):
Suggestions are mostly about accuracy of comments and outcome types. Happy to chat through any of the specific statistics :)

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+  text(xpos, (ypos + 2), sprintf("%.2f", mtc[2]), adj = 0.5, cex = ucex, col = "black")
+  text(xpos, (ypos + 1), sprintf("(%.2f to %.2f)", mtc[1], mtc[3]), adj = 0.5, cex = ucex, col = "black")
+  #PW
+  text(xpos, (ypos - 1), sprintf("%.2f", pw[2]), adj = 0.5, cex = ucex, col = "grey55")
+  if (!is.na(pw[2])) {
+    text(xpos, (ypos - 2), sprintf("(%.2f to %.2f)",  pw[1], pw[3]), adj = 0.5, cex = ucex, col = "grey55")
+  }
+}
+
+#' Function to draw two-tired error bars for summary relative estimates.
```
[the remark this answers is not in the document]
JanionNevill: Fixed

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+  }
+  points(OR + offs, ypos, pch = 15, cex = 0.8 * ulwd, col = ucol, adj = 0.5)
+}
+
+#' Single plot for MTC & MA SFP in NMA SPF Matrix.
+#'
+#' @param mtc Meta-analysis data for direct and indirect evidence
+#' @param pw Meta-analysis data for direct (pairwise) evidence only
+#' @param bpredd TRUE if confidence interval to be plotted as error bars
+#' @param baxis TRUEW if asex to be drawn for forest plots
```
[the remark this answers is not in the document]
JanionNevill: Fixed

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+#' @param scaletype The outcome type being plotted. "HR" for hazard ratio, and "OR" for odds ratio will be plotted on a log scale, anything else will be plotted on a linear scale
```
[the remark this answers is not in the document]
JanionNevill: Changed HR to RR. I got confused which we use

R/summary_forest_plot.R
JanionNevill: Fixed

R/summary_forest_plot.R
JanionNevill: Added handling for other outcomes

R/summary_forest_plot.R
JanionNevill: Corrected

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+  if (!is.na(pw[2])) {
+    text(xpos, (ypos - 2), sprintf("(%.2f to %.2f)",  pw[1], pw[3]), adj = 0.5, cex = ucex, col = "grey55")
+  }
+}
+
+#' Function to draw two-tired error bars for summary relative estimates.
+#'
+#' @param offs Axis offset
+#' @param LL Lower limit
+#' @param OR Point estimate
```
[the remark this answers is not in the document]
JanionNevill: Renamed variables

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+#' @param ypos Position of error bars in Y axis
+#' @param ucol Text colour. Defaults to black
+#' @param ulwd Line width. Defaults to 1
+#' @param pcI  TRUE if both credible interval and predictive interval to be plotted, otherwise only credible interval plotted. Defaults to FALSE
+#' @param predbd Predictive data
+PrICrI <- function(offs, LL, OR, UL, ypos, ucol = "black", ulwd = 1, pcI = FALSE, predbd = c(NA, NA)) {
+  if (pcI) {   #show both CrI & PrI
+    if (predbd[1] != 0 & predbd[2] != 0) {
+      lines(c(predbd[1] + offs, predbd[2] + offs), c(ypos, ypos), lty = 1, lwd = ulwd, col = ucol)  		#pred dist interval
+      lines(c(predbd[1] + offs, predbd[1] + offs), c(ypos - 0.2, ypos + 0.2), lty = 1, lwd = ulwd, col = ucol)  	#pred dist 25% line
```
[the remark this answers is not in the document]
JanionNevill: Updated comments

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+#' @param pw Meta-analysis data for direct (pairwise) evidence only
+#' @param bpredd TRUE if confidence interval to be plotted as error bars
+#' @param baxis TRUEW if asex to be drawn for forest plots
+#' @param scaletype The outcome type being plotted. "HR" for hazard ratio, and "OR" for odds ratio will be plotted on a log scale, anything else will be plotted on a linear scale
+singleSFP <- function(mtc, pw, bpredd = TRUE, baxis = TRUE, scaletype) {
+  
+  ##define axis offset
+  offs = 0
+  
+  #Add reference line; log(1) = 0
```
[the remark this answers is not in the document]
JanionNevill: The line does draw in the correct position for both scale types, but I've removed the needless `log(1)` call

R/summary_forest_plot.R
```diff
@@ -0,0 +1,637 @@
...
+  for (index in 1:ntx) {
+    mtc$rkgram[((index - 1) * ntx) + (1:ntx), 1] <- ranking_matrix[ordered_treatment_names[index], ]
+  }
+  
+  mtc$type <- outcome_type
+  
+  count <- 1
+  for (i in 1:(ntx - 1)) {
+    for (j in (i + 1):ntx) {
+      ma$or[count, 5] <- exp(net1$lower.direct.random[i, j])
```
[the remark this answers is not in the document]
JanionNevill: Added handling for the outcome type. This also now inverts the data correctly for the 2 categories of outcomes

R/summary_forest_plot.R
BalancedmonkeY: The ranking information is still inconsistent. On the plot, the ranks are given as integers (1, 2, 3, then 4), but if it is SUCRA, then the value would be between 0 and 1 (or 0-100%). I think we may need to take a closer look at `mtc$rank[i %% ntx, 3]`

BalancedmonkeY (changes requested):
Code looked okay - went to check the new plot within the app and found the following:

1) Rank value still doesn't quite make sense (see comment on conversation previously discussing this)

2) For at least 6x6 grid, the larger numbers are now writing on top of themselves. Maybe make a dynamic text size that changes size depending on how large the grid is? 

3) The numbers in the SFP do match the other tabs (i.e. they are correct), but they have been placed in the wrong position in the grid (and inverted for some). The symmetry of the values and plots are correct though.

JanionNevill:
Status update: The wrong values are being passed into `singleSFP()` on line 366 in `multiplot()` in `summary_forest_plot.R`. This seems to be because the variable `i.pt` is being used incorrectly. The order of items in the matrix `mtc$lor` does not match the expectation.

The matrices `mtorg` or `mtord` seem to be responsible for the ordering, but I have not been able to discern how to get them to behave as expected. The vector `po` in `mtcMatrixCont()` on line 663 in `summary_forest_plot.R` might also be useful as it defines the ranking order of the treatments. It is used to calculate `mtorg`, but not passed through to the plotting functions.

tommorris168:
Whoever wrote this comment had found the problem:
`# The rankogram command seems to sort the treatments alphabetically first, so this line converts back to the original treatment IDs`
The solution was to sort the output from netmeta by the original treatment order.

I'm sure you'll both agree that the fundamental problem with the code is that the grid is two-dimensional but is treated as one-dimensional, which is why there is a single loop that populates the grid which has opaque lines like this:
`if (i %% (ntx + 1) == 1) {`
...
`} else if (i %% ntx == 0 | i %% ntx > i %/% ntx) {`
...
`} else if (i %% ntx <= i %/% ntx) {`

The output from netmeta is also two-dimensional but is reduced stored in vectors, which is why the easiest place to sort was the netmeta output, rather than later on when it had been collapsed into one dimension.

JanionNevill:
Great job, that looks perfect.

R/summary_forest_plot.R
```diff
@@ -417,7 +437,6 @@ multiplot <- function(stytitle, ntx, lstx, mtc, ma, bpredd = TRUE, plt.adj, ucex
 #'
```
BalancedmonkeY, line 437: Line 411: It would be good if we said somewhere what 'PI' is

BalancedmonkeY (changes requested):
Main issues with plot are now fixed. Please see new comments on some minor issues/suggestions

BalancedmonkeY (approved):
Well done all :)

tommorris168:
@JanionNevill @BalancedmonkeY There have been massive changes to server.R and ui.R since this branch was started, so that resolving the conflicts can't be done in the way I normally do it (which is on Github). I'm thinking the best thing to do is to start a new branch from dev and copy the changes from this branch into that. What do you think?

JanionNevill:
> @JanionNevill @BalancedmonkeY There have been massive changes to server.R and ui.R since this branch was started, so that resolving the conflicts can't be done in the way I normally do it (which is on Github). I'm thinking the best thing to do is to start a new branch from dev and copy the changes from this branch into that. What do you think?

That sounds like a good idea. I had to do that with the disconnected network stuff, when that sat for months without updates.

tommorris168:
Changes from this branch were transferred to another branch. Closing without merging.

PR #86, opened 2023-10-20
Mr 4 placeholder tab

JanionNevill:
Created shell of a tab which is hidden if there are no covariates available. Also fixed issue with data filtering when there are no covariates.


nabury (approved):
From eyeballing the code it all looks good (and works fine without a covariate). Is the plan to add an additional test and manually check when the covariate upload is working?

PR #89, opened 2023-10-24
mr-11-and-13 Non-shiny versions of functions

JanionNevill:
Two example datasets are included. For the purpose of testing the functions it doesn't matter that the models don't converge.


BnmaAnalysis.R
```diff
@@ -0,0 +1,186 @@
...
+                                                 type = effectsType,
+                                                 baseline = covParameters,
+                                                 baseline.risk = "independent"))
+  }
+  if(outcomeType == "continuous"){
+    BRnetwork <- with(BRdata, bnma::network.data(Outcomes = ArmLevel$Outcomes,
+                                                 Study = ArmLevel$Study,                                           
+                                                 Treat = ArmLevel$Treat,
+                                                 response = "normal",
+                                                 SE = ArmLevel$SD,
```
BalancedmonkeY, line 168: Aren't standard error and standard deviation different values? Should there be some transformation?

BnmaAnalysis.R
```diff
@@ -0,0 +1,186 @@
...
+
+
+
+
+
+#' Takes MetaInsight data and converts it into BNMA format
+#'
+#' @param BRdata A data frame of data in MetaInsight format
+#' @param dataFormat "long" or "wide"
+#' @param outcomeType "continuous" or "binomial"
```
BalancedmonkeY, line 114: The options need to be capitalised, i.e. 'Continuous' or 'Binomial', as this is what the shiny input spits out when users select their option.
This relates to lines 128, 130, 152, and 163.

BalancedmonkeY (changes requested):
See comments :)

BnmaAnalysis.R
```diff
@@ -0,0 +1,186 @@
...
+#====================================================================
+#This is roughly the format the data should have. They have used a list to store all variables, whereas I am using a data frame where possible as I think it's clearer.
+bnma::certolizumab
+
+
+#Binomial MetaInsight-long-formatted data
+BRdataBin <- read.csv(".\\Datasets\\BinOutBR.csv")
+
+
+#------------------------------------------
```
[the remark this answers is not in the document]
tommorris168: @BalancedmonkeY Yes, there were plenty of extras in the script to help with your review. They can all be deleted, as per what you wrote on Jira.

BnmaAnalysis.R
tommorris168: @BalancedmonkeY Good point. I think I've had a misunderstanding for a long time about this. In meta-analysis language people generally use variance to mean both SE^2 and SD^2. I have also been using SE and SD interchangeably because I thought they were all SEs, because I thought they were the output from a model rather than raw data. Am I right in saying the y's and their SDs are actually raw data rather than the output from a model? In which case the SDs in MetaInsight data really are SDs and not SEs. In my simulated data I called them SDs but really they are SEs. So the upshot is yes, the function can take SE or the pair (SD, N), and it should be SD and N.

BalancedmonkeY: Yeah, I often get confused between them all as well!
But yeah, it's my understanding that the mean and SD are raw data from each study -> but how each study obtained that information I'm not 100% sure of. 
The fact that on the data upload page of MetaInsight it says "SD contains the standard deviation of the outcome in each arm of the study", I think we can assume there are SD, rather than SE.
So in conclusion, you're change is good :)

BalancedmonkeY (approved):

PR #93, opened 2023-10-31
Mi 29 modularise tabs

JanionNevill:
Split each top-level tab in to a module each. Also removed the privacy notice tab.

Functionality and UI should remain completely unchanged other than the tab removal. I have gone through every tab and checked that everything works, but I could have missed things.


R/data_analysis_page.R
```diff
@@ -0,0 +1,2146 @@
...
+                  ),
+                  column(
+                    width = 6,
+                    numericInput( inputId = ns('label_excluded'), label = 'Label size multiplier', value = 1.25)
+                  ),
+                  downloadButton(outputId = ns('downloadNetworkUpdate')),
+                  br(),
+                  br(),
+                  br(),
+                  verbatimTextOutput("netconnect_sub")
```
nabury, line 232: This text output isn't showing (think I'm in the correct place - tab 1c under the download button for the sensitivity analysis)
Screenshot 2023-10-31 at 11 21 20

R/data_analysis_page.R
```diff
@@ -0,0 +1,2146 @@
...
+                tableOutput(outputId = ns("baye_comparison")),
+                downloadButton(outputId = ns('downloadbaye_comparison')),
+                br(),
+                br(),
+                p(tags$strong("Treatment effects with studies excluded: comparison of all treatment pairs.")),
+                tableOutput(outputId = ns("baye_comparison_sub")),
+                downloadButton(outputId = ns('downloadbaye_comparison_sub'))
+              ),
+              tabPanel(
+                title = "3c. Ranking Panel",
```
nabury, line 454: The relative effects section of the ranking panel is also not displaying (all studies & sensitivity analysis)

nabury (changes requested):
I'll continue to look at this, but I've found a couple of places where elements are missing.

Otherwise, I've tried a random selection of options & download buttons that have all been successful

R/data_analysis_page.R
JanionNevill: Well spotted. I've added the namespacing to that one now.

R/data_analysis_page.R
JanionNevill: I have seen that happen, but I've not been able to reproduce it today. I did find that the plot appeared when the window was resized, so it's not a broken output.

nabury: It's working for me again now ...

nabury (approved):

PR #97, opened 2023-11-07
MR-16 outcome dependent UI

JanionNevill:
Added "informed conditional panel"* to inform a user when a conditional panel is hidden. This is used to hide covariate analysis and baseline risk analysis when the outcome measures are not supported.

Note that I was also going to hide the contents of the bayesian NMA tab, but this would be better done after the modularisation of the data analysis tab.

*Better name suggestions welcome


BalancedmonkeY (commented):
All seems to be working fine.

I would suggest checking with the rest of the team whether we also limit tab 4a to certain outcome measures (plots may only work with certain ones, and we may not want to mislead users).

Bayesian to be completed.

BalancedmonkeY:
> > I would suggest checking with the rest of the team whether we also limit tab 4a to certain outcome measures (plots may only work with certain ones, and we may not want to mislead users).
> 
> I'm talking to @tommorris168 tomorrow about how this tab will work with baseline risk data so it's on my radar. Is there a place where the user will select meta regression/baseline risk as an option before tab 4?

I don't think so - There will be separate tabs for baseline and covariates. I know we discussed on Thursday about whether tab 4a will have both baseline and covariate or a toggle -> can't remember the decision

PR #98, opened 2023-11-07
MR-18 clean treatment labels

JanionNevill:
Cleaned treatment names to work with bayesian and meta-regression

### DIFF
--- a/R/DataWrangling.R
+++ b/R/DataWrangling.R
@@ -31,7 +31,7 @@
 .covariate_prefix_regex <- "^covar\\."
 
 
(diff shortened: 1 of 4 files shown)


BalancedmonkeY (changes requested):
All looks good - thank you

Minor change is to alter the text on the data upload for treatment names to say that any characters that aren't alphabetical, numeric, or underscores will be replaced with an underscore.

JanionNevill:
I've updated the end of project Jira task to include this in the documentation changes.

BalancedmonkeY (approved):

PR #99, opened 2023-11-09
Mi 20 gdpr consent cookie

JanionNevill:


JanionNevill:
GA ID needs to be updated to GA4 ID. Could be generalised to take the ID as a parameter.

server.R
```diff
@@ -11,6 +11,10 @@ shinyServer(function(input, output, session) {
   
   ### GDPR
   
+  google_analytics_header_server(id = "analytics")
+
+  ### View the full update history
+  
   showModal(
```
RyanJField, line 18: This modal needs to be removed, otherwise you get a double warning, which is an issue when the new model is rejected, this one still shows, looking like you have no option but to consent.

ui.R
```diff
@@ -14,7 +14,8 @@ dashboardPage(
       tags$meta(name="keywords", content="MetaInsight, NMA, Network, Meta, Analysis, App"),
       tags$meta(property="og:title", content="Meta Insight: v5.0.1"),
       tags$meta(property="og:description", content="An interactive web tool for network meta-analysis (NMA) that leverages established analysis routines"),
-      tags$meta(property="og:image", content="https://raw.githubusercontent.com/CRSU-Apps/MetaInsight/main/www/images/MetaInsightLogo.png")
+      tags$meta(property="og:image", content="https://raw.githubusercontent.com/CRSU-Apps/MetaInsight/main/www/images/MetaInsightLogo.png"),
+      google_analytics_header_ui(id = "analytics")
     ),
     navbarPage(
       title = "MetaInsight",
```
RyanJField, line 21: The line below (`header...`) needs to be removed otherwise google analytics is loaded all the time.

RyanJField (changes requested):

server.R
JanionNevill: Not sure how I left that in. Removed now.

ui.R
JanionNevill: Removed.

RyanJField (approved):
Looks good!

PR #100, opened 2023-11-10
Mr 1b r roption

BalancedmonkeY:
Added option to choose outcome measure, improved tests, and added LongToWide() (with tests).

Pull requests doesn't have an associated Jira task as this was meant to be part of MR-1 but got forgotten...

### DIFF
--- a/R/DataWrangling.R
+++ b/R/DataWrangling.R
@@ -155,16 +155,16 @@ FindExpectedReferenceTreatment <- function(treatments) {
 #' @param data Data frame in which to search for treatment names
 #' @param treatent_ids Data frame containing treatment names (Label) and IDs (Number)
 #' @return Data frame where the treatments are given as IDs, not names
-ReplaceTreatmentIds <- function(data, treatent_ids) {
+ReplaceTreatmentIds <- function(data, treatment_ids) {
   if (FindDataShape(data) == "long") {
     # Long format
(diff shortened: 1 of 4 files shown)


tommorris168 (changes requested):
Everything works. Some minor changes:

In the headers for LongToWide, WideToLong, PrepDataGemtc, and RunCovariateModel, for the outcome_type parameter, please write binary and continuous as 'Binary' and 'Continuous', i.e. in the exact format.

In LongToWide you haven't done a "complete" if-then-else for Continuous and Binary as per our discussion on Teams. This would also be a good time to do the same for WideToLong!

BalancedmonkeY:
> Everything works. Some minor changes:
> 
> In the headers for LongToWide, WideToLong, PrepDataGemtc, and RunCovariateModel, for the outcome_type parameter, please write binary and continuous as 'Binary' and 'Continuous', i.e. in the exact format.
> 
> In LongToWide you haven't done a "complete" if-then-else for Continuous and Binary as per our discussion on Teams. This would also be a good time to do the same for WideToLong!

Suggestions completed :)

tommorris168 (approved):
I won't point out that in LongToWide there's a missing ' in
`paste0("outcome_type needs to be 'Binary' or 'Continuous")`
because that would be taking things too far ;)

PR #101, opened 2023-11-14
Mi 32 modularise bayesian

JanionNevill:
Modularised tabs 3a. & 3b.


BalancedmonkeY (approved):
After amending sourcing command all worked fine.

PR #102, opened 2023-11-15
Mr 17 improve bnma analysis

tommorris168:
I have made more changes than were in the Jira task, based on gemtc_analysis.


BalancedmonkeY (changes requested):
Changes are only regarding the unit tests (and a reflection on what I learnt from my code review).

1. 5 of the tests failed due to not being able to find CleanTreatmentIds() - so some further rebasing may be needed

2. Unit tests are meant to be quick, so running these regression models many times is undesirable. When I received a code review, the compromise we made was that I could include running the model in the unit tests for the reproducible test - but to alter the others. This was done by breaking the user-written functions up a bit further (see https://github.com/CRSU-Apps/MetaInsight/blob/feature-meta-regression/R/gemtc_analysis.R). You can create a function to produce the network object, then use it in BaselineRiskRegression(). This way, you can do the tests to check assignment of model type etc. just on the network object function (which doesn't take up time running the model). Hope this makes sense :)

BalancedmonkeY (approved):
All tests now passed and much faster

PR #104, opened 2023-11-17
Mr 20 covariate value input

JanionNevill:
Added inputs to control covariate value in analysis.

### DIFF
--- a/R/covariate_analysis_panel.R
+++ b/R/covariate_analysis_panel.R
@@ -11,35 +11,51 @@ covariate_analysis_panel_ui <- function(id) {
         style = "display: inline-block; vertical-align: top; padding-right: 20pt;"
       ),
       div(
+        # Type selection if the covariate data is valid
         conditionalPanel(
           condition = "output.valid_covariate",
           ns = ns,
-          selectInput(
-            inputId = ns("covariate_type_selection"),
-            label = "",
-            choices = c("Binary", "Continuous"),
-            width = "120pt"
+          div(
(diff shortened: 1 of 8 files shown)


R/covariate_analysis_panel.R
```diff
@@ -20,14 +17,15 @@ covariate_analysis_panel_ui <- function(id) {
...
+            choices = c("Binary", "Continuous"),
+            width = "120pt"
           )
         ),
         style = "display: inline-block;"
       ),
       div(
         conditionalPanel(
-          condition = "output.inferred_type == 'continuous'",
+          condition = "output.inferred_type == 'Continuous'",
```
nabury, line 28: I'm confused about the logic behind this message "If your data is binary, the only allowed values are 0, 1, and NA"

nabury (changes requested):
Could this all have some comments please? It's taking me a long time to work out what is actually happening in each section.

R/covariate_analysis_panel.R
JanionNevill: I've changed the wording to hopefully make it clearer. Happy to workshop it if you think that it still needs it.

nabury: > I've changed the wording to hopefully make it clearer. Happy to workshop it if you think that it still needs it.

That's much clearer

nabury (approved):
Comments make it much easier to follow the logic of the code. All seems to work correctly and tests passed. Changed a minor typo.

PR #110, opened 2023-11-29
MI-37 modularise bayesian

JanionNevill:
Modularised the rest of the bayesian tabs.  I still need to check that all of the download buttons still work correctly.


BalancedmonkeY (approved):
All the download plots seemed to work as well

PR #116, opened 2023-12-18
MR 25 Contribution matrix and unit tests.

tommorris168:
Everything is ready except for testing the contribution matrix itself. I have one example where we know what the output is supposed to be, and another bigger one where I might have to type in the data myself because it's in a PDF. The two examples only cover one case out of six, so for the other five we might only be able to check that the matrix is not too different than in the first case. Based on my testing so far, fixed or random effects can change the matrix a fair bit but shared/exchangeable/unrelated doesn't seem to.


BalancedmonkeY:
I've asked Janion for a review to see if there are any improvements that can be made regarding matrix/vector calculations/wrangling :)

R/DataWrangling.R
```diff
@@ -285,3 +285,51 @@ FindCovariateNames <- function(df) {
...
+  #When used in tapply it matches the treatments within a study to the ordered 'treatments' vector, and then finds the lowest
+  min_match <- function(x){min(match(x, treatments))}
+  #Find the control treatment in each study
+  control <- data.frame(Study = studies, Control = treatments[tapply(data$Treatment, INDEX = data$Study, FUN = min_match)])
+  data <- merge(data, control, by = "Study", sort = FALSE)
+  if(keep_delete == "keep"){
+    return(data[data$Treatment == data$Control, ])
+  } else if(keep_delete == "delete"){
+    return(data[data$Treatment != data$Control, ])
+  } else{stop("keep_delete must be 'keep' or 'delete'")}
```
JanionNevill, line 312: Please move this onto an indented block

R/DataWrangling.R
```diff
@@ -285,3 +285,51 @@ FindCovariateNames <- function(df) {
...
+#' @param keep_delete "keep" or "delete".
+#' @return @param data with rows corresponding to the control treatment kept or deleted, and a new column 'Control'.
+KeepOrDeleteControlTreatment <- function(data, treatments, keep_delete){
+  #It is imperative that 'data' is sorted by 'Study', because tapply below will sort by 'Study', creating a misalignment between 'Study' and 'Control' in the 'control' data frame if 'data' is not sorted by 'Study' already.
+  data <- dplyr::arrange(data, Study, T)
+  #The unique studies
+  studies <- unique(data$Study)
+  #Local function to find the control treatment in a single study
+  #When used in tapply it matches the treatments within a study to the ordered 'treatments' vector, and then finds the lowest
+  min_match <- function(x){min(match(x, treatments))}
```
JanionNevill, line 304: Can you make this an indented block please? It makes it a bit clearer

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+  #Reference treatment
+  reference <- treatments[1]
+  #Data with control treatment rows deleted
+  data_no_control <- DeleteControlTreatment(data = data, treatments = treatments)
+  #Used as labels for the matrix rows
+  study_treatment <- paste0("(", data_no_control$Study, ")", data_no_control$Control, ":", data_no_control$Treatment)
+  
+  #Create the design matrix with the right dimensions and zeros everywhere
+  if(cov_parameters %in% c("Unrelated", "Exchangeable")){
+    X <- matrix(0, nrow = length(study_treatment), ncol = 2 * (n_treatments - 1))
```
JanionNevill, line 110: Single-letter variable names are discouraged because it's difficult to identify what they are meant to represent. It'd fine in a 1-line function, but here a more explicit name would be helpful

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+        #...put 1 in the treatment column...
+        X[which(GetStudies(rownames(X)) == data_no_control$Study[i] & GetTreatments(rownames(X)) == data_no_control$Treatment[i]), which(GetTreatments(colnames(X)) == data_no_control$Treatment[i])] <- 1
+        #...put -1 in the control treatment column...
+        X[which(GetStudies(rownames(X)) == data_no_control$Study[i] & GetTreatments(rownames(X)) == data_no_control$Treatment[i]), which(GetTreatments(colnames(X)) == data_no_control$Control[i])] <- -1
+        #...put the covariate value in the treatment's covariate column...
+        X[which(GetStudies(rownames(X)) == data_no_control$Study[i] & GetTreatments(rownames(X)) == data_no_control$Treatment[i]), which(GetTreatments(colnames(X)) == paste0(data_no_control$Treatment[i], "_beta"))] <- covar_centered[data_no_control$Study[i]]
+        #...and put minus the covariate value in the control treatment's covariate column
+        X[which(GetStudies(rownames(X)) == data_no_control$Study[i] & GetTreatments(rownames(X)) == data_no_control$Treatment[i]), which(GetTreatments(colnames(X)) == paste0(data_no_control$Control[i], "_beta"))] <- -covar_centered[data_no_control$Study[i]]
```
JanionNevill, line 149: These blocks look like they're doing the same calculations multiple times to get the matrix indices. Can the index calculations be brought out into variables before the `if` statements, then just used in these lines?

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+        X[which(GetStudies(rownames(X)) == data_no_control$Study[i] & GetTreatments(rownames(X)) == data_no_control$Treatment[i]), which(GetTreatments(colnames(X)) == "B")] <- covar_centered[data_no_control$Study[i]]
+        
+        #If the study does not contain the reference treatment...
+      } else if(data_no_control$Control[i] != reference){
+        #...put 1 in the treatment column...
+        X[which(GetStudies(rownames(X)) == data_no_control$Study[i] & GetTreatments(rownames(X)) == data_no_control$Treatment[i]), which(GetTreatments(colnames(X)) == data_no_control$Treatment[i])] <- 1
+        #...put -1 in the control treatment column...
+        X[which(GetStudies(rownames(X)) == data_no_control$Study[i] & GetTreatments(rownames(X)) == data_no_control$Treatment[i]), which(GetTreatments(colnames(X)) == data_no_control$Control[i])] <- -1
+        #...and put 0 in the covariate column
+        X[which(GetStudies(rownames(X)) == data_no_control$Study[i] & GetTreatments(rownames(X)) == data_no_control$Treatment[i]), which(GetTreatments(colnames(X)) == "B")] <- 0
```
JanionNevill, line 171: Similar to the above: calculations don't need to be repeated.

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+      }
+    }
+  }
+  
+  return(X)
+}
+
+
+
+#' Create the Z matrix.
```
JanionNevill, line 181: What's a "Z matrix"? Could the roxygen description be expanded please?

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+    second_treatment <- substr(rownames(Z_top_left)[i],
+                               start = unlist(gregexpr(":", rownames(Z_top_left)[i])) + 1,
+                               stop = nchar(rownames(Z_top_left)[i]))
+    #If the first treatment in the row is the reference...
+    if(first_treatment == reference){
+      #...put 1 in the column correseponding to the second treatment
+      Z_top_left[i, which(colnames(Z_top_left) == second_treatment)] <- 1
+    }
+    #If the first treatment in the row is not the reference...
+    else if(first_treatment != reference){
```
JanionNevill, line 223: This can just be an `else`, because the 2 conditions are exact opposites

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+                     cov_parameters = cov_parameters)
+  Z <- CreateZMatrix(treatments = treatments, cov_parameters = cov_parameters)
+  
+  if(effects_type == "Fixed"){
+    
+    if(cov_parameters %in% c("Unrelated", "Shared")){
+      
+      if(all_parameters){
+        contribution <- Z %*% solve(t(X) %*% solve(V) %*% X) %*% t(X) %*% solve(V)
+      } else if(!all_parameters){
```
JanionNevill, line 370: This can just be an `else`

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+        contribution <- solve(t(X) %*% solve(V) %*% X) %*% t(X) %*% solve(V)
+      }
+      
+    } else if(cov_parameters == "Exchangeable"){
+      
+      if(is.null(std_dev_beta)){stop("Must specify 'std_dev_beta' when cov_parameters == 'Exchangeable'")}
```
JanionNevill, line 376: Please make into an indented block

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+      if(all_parameters){
+        contribution <- Z %*% rbind(A_top_left, A_middle_left)
+      } else if(!all_parameters){
+        contribution <- rbind(A_top_left, A_middle_left)
+      }
+    }
+    
+  } else if(effects_type == "Random"){
+    
+    if(is.null(std_dev_d)){stop("Must specify 'std_dev_d' when effects_type == 'Random'")}
```
JanionNevill, line 418: Please move into an indented block

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+      #Populate the bottom right of X_star
+      X_star[(nrow(X) + 1):nrow(X_star), (ncol(X_d) + ncol(X_beta) + 1):ncol(X_star)] <- -matrix(1, nrow = ncol(X_d), ncol = 1)
+      
+      A <- solve(t(X_star) %*% solve(V_star) %*% X_star) %*% t(X_star) %*% solve(V_star)
+      A_top_left <- A[1:ncol(X_d), 1:nrow(X_d)]
+      A_middle_left <- A[(ncol(X_d) + 1):(ncol(X_d) + ncol(X_beta)), 1:nrow(X_beta)]
+      
+      if(all_parameters){
+        contribution <- Z %*% rbind(A_top_left, A_middle_left)
+      } else if(!all_parameters){
```
JanionNevill, line 411: This be an `else`

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+      X_star[(nrow(X) + 1):nrow(X_star), 1:nrow(X)] <- diag(1, nrow = nrow(X))
+      #Populate the bottom right of X_star
+      X_star[(nrow(X) + 1):nrow(X_star), (nrow(X) + 1):ncol(X_star)] <- -X
+      
+      A <- solve(t(X_star) %*% solve(V_star) %*% X_star) %*% t(X_star) %*% solve(V_star)
+      A_bottom_left <- A[(nrow(X) + 1):nrow(A), 1:nrow(X)]
+      
+      if(all_parameters){
+        contribution <- Z %*% A_bottom_left
+      } else if(!all_parameters){
```
JanionNevill, line 447: This can be an `else`

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+        contribution <- A_bottom_left
+      }
+      
+    } else if(cov_parameters == "Exchangeable"){
+      
+      if(is.null(std_dev_beta)){stop("Must specify 'std_dev_beta' when cov_parameters == 'Exchangeable'")}
```
JanionNevill, line 453: Please move into an indented block

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+      #Populate the bottom right column of X_star
+      X_star[(2 * nrow(X) + 1):nrow(X_star), (nrow(X) + 2 * ncol(X_d) + 1):ncol(X_star)] <- -matrix(1, nrow = ncol(X_d), ncol = 1)
+      
+      A <- solve(t(X_star) %*% solve(V_star) %*% X_star) %*% t(X_star) %*% solve(V_star)
+      A_row2_left <- A[(ncol(X) + 1):(ncol(X) + ncol(X_d)), 1:nrow(X)]
+      A_row3_left <- A[(ncol(X) + ncol(X_d) + 1):(ncol(X) + 2 * ncol(X_d)), 1:nrow(X)]
+      
+      if(all_parameters){
+        contribution <- Z %*% rbind(A_row2_left, A_row3_left)
+      } else if(!all_parameters){
```
JanionNevill, line 494: This can be an `else`

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+  }
+  
+  contribution_abs <- abs(contribution)
+  contribution_row_sums <- rowSums(contribution_abs)
+  contribution_percent <- 100 * contribution_abs / contribution_row_sums
+  
+  #Select percentage contributions or absolute contributions
+  if(percentages){
+    contribution_output <- t(contribution_percent)
+  } else if(!percentages){
```
JanionNevill, line 510: This can be an `else`

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+  }
+  
+  #For some configurations the column names are not carried through, so put them back in
+  if(is.null(colnames(contribution_output))){
+    colnames(contribution_output) <- colnames(X)
+  }
+  
+  if(!full_output){
+    return(round(contribution_output, digits = 2))
+  } else if(full_output){
```
JanionNevill, line 551: This can be an `else`

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+#'           The contribution matrix.
+#'         If @param full_output = TRUE:
+#'           List
+#'            - 'contribution' = contribution matrix.
+#'            - 'V' = The treatment effect variance matrix.
+#'            - 'X' = The design matrix.
+#'            - 'Z' = The Z matrix.
+#'            - 'Lambda_tau' = The Lambda_tau matrix (only included if @param effects_type == "Random").
+#'            - 'Lambda_beta' = The Lambda_beta matrix (only included if @param cov_parameters == "Exchangeable").
+CreateContributionMatrix <- function(data, treatment_ids, outcome_type, effects_type, std_dev_d = NULL, cov_parameters, std_dev_beta = NULL, study_level, percentages, all_parameters = TRUE, full_output = FALSE){
```
JanionNevill, line 341: This function is ~250 lines long, so it would be good to break it up. I'm not sure of the best division, but there seem to be smaller chunks which could be factored out.

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+      #Populate the middle left of X_star
+      X_star[(nrow(X) + 1):(2 * nrow(X)), 1:nrow(X)] <- diag(1, nrow = nrow(X))
+      #Populate the middle second column of X_star
+      X_star[(nrow(X) + 1):(2 * nrow(X)), (nrow(X) + 1):(nrow(X) + ncol(X_d))] <- -X_d
+      #Populate the middle third column of X_star
+      X_star[(nrow(X) + 1):(2 * nrow(X)), (nrow(X) + ncol(X_d) + 1):(nrow(X) + 2 * ncol(X_d))] <- -X_beta
+      #Populate the bottom third column of X_star
+      X_star[(2 * nrow(X) + 1):nrow(X_star), (nrow(X) + ncol(X_d) + 1):(nrow(X) + 2 * ncol(X_d))] <- diag(1, nrow = ncol(X_d))
+      #Populate the bottom right column of X_star
+      X_star[(2 * nrow(X) + 1):nrow(X_star), (nrow(X) + 2 * ncol(X_d) + 1):ncol(X_star)] <- -matrix(1, nrow = ncol(X_d), ncol = 1)
```
JanionNevill, line 486: Are any of these recalculated indices which could be calculated once and stored in a variable?

tests/testthat/test-contribution_matrix.R
```diff
@@ -0,0 +1,244 @@
+
+test_that("CreateVMatrix() works for binary outcomes", {
+  data <- data.frame(Study = c("A", "A", "B", "B", "C", "C", "C", "D", "D"),
```
JanionNevill, line 3: This data frame is used in multiple test cases, could it be extracted into a `.csv` file and read in at the beginning of each case? That way any future changes to the data set can be done in one central location.

tests/testthat/test-contribution_matrix.R
```diff
@@ -0,0 +1,244 @@
...
+  outcome_type <- "Binary"
+
+  #Variance formula from "Introducion to Meta-Analysis" 2nd edition, Borenstein et al, page 36.
+  effect_variances <- vector(length=5)
+  effect_variances[1] <- 1/8 + 1/(40-8) + 1/13 + 1/(41-13)
+  effect_variances[2] <- 1/14 + 1/(42-14) + 1/21 + 1/(43-21)
+  effect_variances[3] <- 1/10 + 1/(44-10) + 1/15 + 1/(45-15)
+  effect_variances[4] <- 1/10 + 1/(44-10) + 1/22 + 1/(46-22)
+  effect_variances[5] <- 1/23 + 1/(47-23) + 1/30 + 1/(48-30)
+  V_matrix <- diag(effect_variances)
```
JanionNevill, line 20: Can you change the names in all of these test cases to be something like `expected_v_matrix` please? Again, it just helps with readability.

JanionNevill (changes requested):
Mostly small formatting improvements

R/DataWrangling.R
```diff
@@ -285,3 +285,51 @@ FindCovariateNames <- function(df) {
...
+  if(keep_delete == "keep"){
+    return(data[data$Treatment == data$Control, ])
+  } else if(keep_delete == "delete"){
+    return(data[data$Treatment != data$Control, ])
+  } else{stop("keep_delete must be 'keep' or 'delete'")}
+}
+
+
+
+#' Delete rows in @param data corresponding to the control treatment in each study.
```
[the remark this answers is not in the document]
tommorris168: @JanionNevill What do you think about this? I know what I've done is a standard thing to do in software development, but I'm not sure why.

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+  #Number of studies
+  n_studies <- length(studies)
+  #Data with only control treatment rows kept
+  data_control <- KeepControlTreatment(data = data, treatments = treatments)
+  #Data with control treatment rows deleted
+  data_no_control <- DeleteControlTreatment(data = data, treatments = treatments)
+  #Create a wide version of data
+  data_wide <- merge(data_control, data_no_control, by = "Study")
+  #Used as labels for the matrix rows and columns
+  study_treatment <- paste0("(", data_no_control$Study, ")", data_no_control$Control, ":", data_no_control$Treatment)
```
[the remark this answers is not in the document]
tommorris168: Done. There is also a variable named `treat_treat` later on that I renamed `treat1_treat2_label`.

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+}
+
+
+
+#' Create the design matrix.
+#' 
+#' @param data Input data in long format plus the column 'Treatment', a text version of 'T'.
+#' @param studies Vector of studies.
+#' @param treatments Vector of treatments with the reference treatment first.
+#' @param covar_centered Vector of centered covariate values, named by study.
```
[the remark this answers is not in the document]
tommorris168: I have added an argument `cov_centre` to CreateContributionMatrix() which allows the centring value to be set, defaulting to the mean.

R/contribution_matrix.R
tommorris168: See comment about the X matrix.

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+#' @param data Input data in long format.
+#' @param treatment_ids Data frame containing treatment IDs and names in columns named 'Number' and 'Label' respectively.
+#' @param outcome_type "Continuous" or "Binary".
+#' @param effects_type "Fixed" or "Random".
+#' @param std_dev_d Between-study standard deviation. Only required when @param effects_type == "Random". Defaults to NULL.
+#' @param cov_parameters "Shared", "Exchangeable", or "Unrelated".
+#' @param std_dev_beta Standard deviation of covariate parameters. Only required when @param cov_parameters == "Exchangeable". Defaults to NULL.
+#' @param study_level TRUE for study-level contributions, FALSE for basic-comparison-level contributions.
+#' @param percentages TRUE for percentage contributions, FALSE for absolute contributions.
+#' @param all_parameters TRUE for one column per parameter, FALSE for one column per basic parameter. Defaults to TRUE.
```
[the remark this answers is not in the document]
tommorris168: Changed.

R/contribution_matrix.R
tommorris168: Created four smaller functions to be called from within this function.

R/contribution_matrix.R
tommorris168: I have gone further and created an index for all of them, even those that are used only once. It looks neater.

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+      X_star[(nrow(X) + 1):(2 * nrow(X)), (nrow(X) + 1):(nrow(X) + ncol(X_d))] <- -X_d
+      #Populate the middle third column of X_star
+      X_star[(nrow(X) + 1):(2 * nrow(X)), (nrow(X) + ncol(X_d) + 1):(nrow(X) + 2 * ncol(X_d))] <- -X_beta
+      #Populate the bottom third column of X_star
+      X_star[(2 * nrow(X) + 1):nrow(X_star), (nrow(X) + ncol(X_d) + 1):(nrow(X) + 2 * ncol(X_d))] <- diag(1, nrow = ncol(X_d))
+      #Populate the bottom right column of X_star
+      X_star[(2 * nrow(X) + 1):nrow(X_star), (nrow(X) + 2 * ncol(X_d) + 1):ncol(X_star)] <- -matrix(1, nrow = ncol(X_d), ncol = 1)
+      
+      A <- solve(t(X_star) %*% solve(V_star) %*% X_star) %*% t(X_star) %*% solve(V_star)
+      A_row2_left <- A[(ncol(X) + 1):(ncol(X) + ncol(X_d)), 1:nrow(X)]
```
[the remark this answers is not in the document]
tommorris168: Good spot! The program is correct, the mistake is in the document. The number of rows to ignore is the number of deltas, which is ncol(X) = 2(T-1). I have fixed this as well as the number-of-studies mistake in the beginning.

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+    stop("effects_type must be 'Fixed' or 'Random'")
+  }
+  
+  contribution_abs <- abs(contribution)
+  contribution_row_sums <- rowSums(contribution_abs)
+  contribution_percent <- 100 * contribution_abs / contribution_row_sums
+  
+  #Select percentage contributions or absolute contributions
+  if(percentages){
+    contribution_output <- t(contribution_percent)
```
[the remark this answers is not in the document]
tommorris168: No, it's just what they did in the Donegan paper.

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+    contribution_study <- matrix(nrow = n_studies, ncol = ncol(contribution_output))
+    rownames(contribution_study) <- studies
+    colnames(contribution_study) <- colnames(contribution_output)
+    for(study in studies){
+      #If there is only one row in the contribution matrix for this study, copy it
+      if(n_arms[study] == 2){
+        contribution_study[which(rownames(contribution_study) == study), ] <- contribution_output[which(rownames(contribution_output) == study), ]
+      #If there is more than one row for this study, add up the rows
+      } else if(n_arms[study] > 2){
+        contribution_study[which(rownames(contribution_study) == study), ] <- colSums(contribution_output[which(rownames(contribution_output) == study), ])
```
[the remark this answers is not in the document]
tommorris168: I don't know what you mean. Can you explain further?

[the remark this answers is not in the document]
BalancedmonkeY: Say somebody wanted a contribution matrix at the study level made up of percentages where there are some studies with more than 2 arms.
From what I can see, first the percentages are calculated across each comparison. Then, if the user has selected 'study level', for studies that had multiple arms, the percentages are summed (column wise). Naturally this is fine for absolute values, but for percentages that would mean that they would add up to more than 100% (i.e. it would be 200% for 3-arm studies, 300% for 4-arm studies...)?
I think if it did the study level thing first (i.e. summed via columns for studies with multiple arms) and then did the percentage thing it would be fine. But with the way round it currently is, I think you'd need to divide the percentages by (number of arms - 1).

BalancedmonkeY (changes requested):
I've explained a bit more about what I meant on the study-level percentages contribution calculations.

Also, when I ran the unit test, 3 of them failed:
CreateContributionMatrix() produces a matrix of the correct format for all three covariate parameter assumptions
CreateContributionMatrix() produces a matrix of the correct format when study_or_comparison_level = 'study'
CreateContributionMatrix() produces a matrix of the correct format when basic_or_all_parameters = 'basic'

The error was the same for all of them:
`Error in ``$<-.data.frame`(`*tmp*`, Treatment, value = character(0))`: replacement has 0 rows, data has 9`

R/DataWrangling.R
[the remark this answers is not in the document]
JanionNevill: I could go either way on this. Generally when this pattern is used, it because the intermediate functions do something which is less obvious. In this case, they just add a string which matches the name, so it's not really adding any more clarity. I think for that reason, there isn't much point having the 2 intermediate functions.

R/contribution_matrix.R
JanionNevill: Ok, not ideal, but makes sense here.

R/contribution_matrix.R
JanionNevill: This wasn't addressed in recent commits

tests/testthat/test-contribution_matrix.R
JanionNevill: Is there a reason that the file with a covariate has `;` as the delimiter instead of `,`?

JanionNevill (changes requested):
Just tiny final comments

R/contribution_matrix.R
```diff
@@ -0,0 +1,588 @@
...
+  #Data with control treatment rows deleted
+  data_no_control <- DeleteControlTreatment(data = data, treatments = treatments)
+  #Create a wide version of data
+  data_wide <- merge(data_control, data_no_control, by = "Study")
+  #Used as labels for the matrix rows and columns
+  study_treatment <- paste0("(", data_no_control$Study, ")", data_no_control$Control, ":", data_no_control$Treatment)
+  
+  if(outcome_type == "Binary"){
+    #Calculate treatment effect variances
+    effect_sizes <- metafor::escalc(measure = "OR",
```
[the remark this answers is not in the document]
tommorris168: Done

R/DataWrangling.R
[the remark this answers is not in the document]
tommorris168: Deleted.

tests/testthat/test-contribution_matrix.R
tommorris168: It seems to be the default in Open Office and I forgot to change it. Changed.

tommorris168:
> When I ran the unit test, 3 of them failed: CreateContributionMatrix() produces a matrix of the correct format for all three covariate parameter assumptions CreateContributionMatrix() produces a matrix of the correct format when study_or_comparison_level = 'study' CreateContributionMatrix() produces a matrix of the correct format when basic_or_all_parameters = 'basic'
> 
> The error was the same for all of them: `` Error in ``$<-.data.frame ``(`*tmp*`, Treatment, value = character(0))`: replacement has 0 rows, data has 9`

It looks like this was caused by the csv file being delimited with ; rather than , as per Janion's comment. Fixed.

R/contribution_matrix.R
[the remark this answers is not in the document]
tommorris168: I think the confusion here is because the matrix is transposed (somewhat arbitrarily as per a previous comment). The percentages are by row (meaning each row sums to 100) before transposing. After transposing the percentages are by column. So the crucial thing is whether the study level matrix is created before or after transposing. Line 680 is where the percentage matrix is transposed, before the study level matrix is created. Do we agree now?

JanionNevill (approved):

R/contribution_matrix.R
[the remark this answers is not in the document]
BalancedmonkeY: Thanks for the clarification. I turns out that I got confused with essentially what the column headers were for the contribution matrix (before transposition) and though they were already at the study level, and so thought we were adding stuff together in a different dimension.

Anyway - I'm happy now - thanks :)

BalancedmonkeY (approved):

PR #117, opened 2023-12-21
MR-24 Regression plot page

JanionNevill:
Added regression plot page.

### DIFF
--- a/R/covariate_analysis_panel.R
+++ b/R/covariate_analysis_panel.R
@@ -36,7 +36,7 @@ covariate_analysis_panel_ui <- function(id) {
                 style = "color: orange;"
               )
             ),
-            style = "display: inline-block; vertical-align: 50%"
+            style = "display: inline-block; vertical-align: 65%"
           ),
           div(
             covariate_value_panel_ui(id = ns("covariate_value")),
@@ -67,6 +67,10 @@ covariate_analysis_panel_ui <- function(id) {
               align = "center",
               bayesian_forest_plot_plus_stats_ui(id = ns("cov_forest_plots"))
             )
+          ),
+          tabPanel(
+            title = "4c-2. Regression plot",
+            regression_plot_panel_ui(id = ns("regression_plot"))
           )
         )
       )
@@ -78,7 +82,8 @@ covariate_analysis_panel_ui <- function(id) {
 #'
 #' @param id ID of the module
 #' @param all_data Study data including covariate columns, in wide or long format
-#' @param treatment_df Reactive containing data frame containing treatment IDs (Number) and names (Label)
+#' @param treatment_df Reactive containing data frame containing treatment IDs (Number), sanitised names (Label), and original names (RawLabel)
+#' @param reference_treatment Reactive containing the sanitised name of the reference treatment
 #' @param metaoutcome Reactive containing meta analysis outcome: "Continuous" or "Binary"
 #' @param outcome_measure Reactive containing meta analysis outcome measure: "MD", "SMD", "OR, "RR", or "RD"
 #' @param model_effects Reactive containing model effects: either "random" or "fixed"
@@ -87,6 +92,7 @@ covariate_analysis_panel_server <- function(
     id, 
     all_data,
     treatment_df,
+    reference_treatment,
     metaoutcome,
     outcome_measure,
     model_effects,
@@ -161,6 +167,7 @@ covariate_analysis_panel_server <- function(
       covariate_type = reactive({ input$covariate_type_selection }),
       covariate_data = reactive({ all_data()[[covariate_title()]] })
     )
+    
     # obtain gemtc output types to be used in rest of page
     model_output <- reactive(CovariateModelOutput(model = model_reactive(), cov_value = covariate_value()))
     # Create forest plot and associated statistics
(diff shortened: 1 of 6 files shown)


R/meta_regression/regression_plot_panel.R
```diff
@@ -0,0 +1,276 @@
...
+      actionButton(inputId = ns("remove_comparator_btn"), label = "Remove"),
+      div(
+        style = "float: right;",
+        actionButton(inputId = ns("remove_all_btn"), label = "Remove all")
+      ),
+      
+      h3("Plot Options"),
+      # Covariate value
+      .AddTooltip(
+        checkboxInput(inputId = ns("covariate"), label = "Show covariate value", value = TRUE),
```
nabury, line 54: I really like the tooltips but it isn't obvious that they are there. Is there any way of making it more obvious without having a messy looking ui?

R/meta_regression/regression_plot_panel.R
```diff
@@ -0,0 +1,276 @@
...
+      .AddTooltip(
+        checkboxInput(inputId = ns("contributions"), label = "Show contributions", value = TRUE),
+        tooltip = "Show study contributions as circles, where a bigger circle represents a larger contribution"
+      ),
+      
+      div(
+        id = ns("contribution_options"),
+        p("Study circle sized by:"),
+        div(
+          .CreateInlineBlock("% Contribution", style = "padding-right: 10pt;"),
```
nabury, line 89: The labelling for this "jumps around" on smaller screen widths
Screenshot 2023-12-21 at 14 21 34

R/meta_regression/regression_plot_panel.R
```diff
@@ -0,0 +1,276 @@
...
+      ),
+      # Contributions
+      .AddTooltip(
+        checkboxInput(inputId = ns("contributions"), label = "Show contributions", value = TRUE),
+        tooltip = "Show study contributions as circles, where a bigger circle represents a larger contribution"
+      ),
+      
+      div(
+        id = ns("contribution_options"),
+        p("Study circle sized by:"),
```
nabury, line 87: This doesn't get "ghosted out" like the slider which makes it seem like it's in use all the time

R/meta_regression/regression_plot_panel.R
```diff
@@ -0,0 +1,276 @@
...
+          .CreateInlineBlock("% Contribution", style = "padding-right: 10pt;"),
+          shinyWidgets::materialSwitch(
+            inputId = ns("contribution_toggle"),
+            inline = TRUE,
+            status = "primary"
+          ),
+          .CreateInlineBlock("Inverse Variance")
+        ),
+        div(
+          .CreateInlineBlock(
```
nabury, line 98: I like the disabling of the widgets but here, where there are two linked to the "show contributions" tick box it is less obvious that the circle size multiplier is linked

nabury (changes requested):
The 'study circle sized by' widget needs fixing to display properly on narrower screens as it's confusing (a shinyWidgets radio group button could be an option so that the text is displayed on the switch rather than beside it).

I don't find the linking of the two widgets that are disabled by the 'show contributions' selector really intuitive/obvious but I suspect it's one of those things that everyone would have an opinion on and it would be difficult to get a perfect solution.

R/meta_regression/regression_plot_panel.R
JanionNevill: I've changed these to both be standard inputs, so they behave better.

R/meta_regression/regression_plot_panel.R
JanionNevill: That's a limitation of the widget. I've changed it to be a standard input, so it behaves now.

R/meta_regression/regression_plot_panel.R
JanionNevill: It should be more obvious now that all the disabling is fixed.

R/meta_regression/covariate_value_panel.R
```diff
@@ -14,43 +14,50 @@
...
-      ns = ns,
-      .CreateInlineBlock(
-        numericInput(
-          inputId = ns("numeric"),
-          label = NULL,
-          min = -.Machine$double.xmax,
-          max = .Machine$double.xmax,
-          value = 0,
-          width = "100pt"
+    title = "This value will be used by all regression analysis output",
```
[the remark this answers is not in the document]
JanionNevill: This is intended to be here as it applies to all of the tabs.

R/covariate_analysis_panel.R
```diff
@@ -36,7 +36,7 @@ covariate_analysis_panel_ui <- function(id) {
                 style = "color: orange;"
```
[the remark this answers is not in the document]
JanionNevill: The click menus are good for when there is a lot of information,. or formatted text. When the information is minimal and all on one line, a standard tooltip is generally preferred.

R/meta_regression/regression_plot_panel.R
JanionNevill: I've added tooltip icons to make it a little more clear.

nabury (approved):
I like the changes. Fixed a typo I spotted.

PR #125, opened 2024-01-12
Mr 6 data plot

nabury:
Includes:

- summary panel 4a
- summary characteristic plot in panel 4a that displays baseline risk or the covariate (when there is one)
- plot download buttons

Known issues:
- there is currently an occasional error message that displays for <1 s when the plot is rendered and I'm not sure how to prevent this 
- the plot may not display _optimally_ i.e. appears "squashed" or large plots disappear off the side of the screen. This also includes plot downloads. Please ignore this for now as it will be addressed in a future task


R/SummaryPanel4a.R
```diff
@@ -0,0 +1,67 @@
+#' Module UI for the covariate summary plot tab.
+#' 
+#' @param id ID of the module
+#' @return Tab 4a Summary
+metaregression_summary_panel_ui <- function(id) {
+  ns <- NS(id)
+  fluidPage(
+    h4('Summary Characteristic Plot'),
+    uiOutput(ns('toggle_covariate_baseline')),
```
JanionNevill, line 9: Can you make this a `shinyWidgets::radioGroupButtons` then use `shinyWidgets::updateRadioGroupButtons()` in the server please? It's best to have as few `uiOutput`s as possible because it pollutes the server logic.

R/SummaryPanel4a.R
```diff
@@ -0,0 +1,67 @@
...
+  )
+}
+
+#' Module server for the covariate summary plot tab.
+#'
+#' @param id ID of the module
+#' @param all_data Study data including covariate columns, in wide or long format
+#' @param metaoutcome Reactive containing meta analysis outcome: 'Continuous' or 'Binary'
+#' @param treatment_df Reactive containing data frame containing treatment IDs (Number) and names (Label)
+#' @return Covariate summary plot tab
```
JanionNevill, line 26: I don't think that there is a return from this function, so this can be removed from the roxygen.

R/SummaryPanel4a.R
```diff
@@ -0,0 +1,67 @@
...
+        shinyWidgets::radioGroupButtons(
+              inputId = session$ns('toggle_covariate_baseline'),
+              choices = c('Covariate', 'Baseline risk'),
+              status = 'primary'
+        )
+      }
+    })
+    
+    # Render covariate summary plot
+    observe({
```
JanionNevill, line 44: The `observe()` wrapper is not needed here. The `renderPlot()` call takes care of all of the reactivity.

R/create_covariate_summary_plot.R
```diff
@@ -0,0 +1,158 @@
+#' Create the covariate summary plot 
+#' https://rdrr.io/github/audrey-b/BUGSnet/man/data.plot.html
+#'
+#' @param all_data Study data including covariate columns, in wide or long format
+#' @param metaoutcome Reactive containing meta analysis outcome: "Continuous" or "Binary"
+#' @param covariate_or_baseline Output from widget to toggle between covariate and baseline risk for the plot type.
+#' @param treatment_df Reactive containing data frame containing treatment IDs (Number) and names (Label)
```
JanionNevill, line 7: These parameters (metaoutcome & treatment_df) aren't reactives, they're the data itself.

R/create_covariate_summary_plot.R
```diff
@@ -0,0 +1,158 @@
...
+#' @return BUGSnet::data.plot plot
+
+CreateCovariateSummaryPlot <- function(all_data, metaoutcome, covariate_or_baseline, treatment_df) {
```
JanionNevill, line 10: This function is 128 lines long, can it be broken up into smaller functions please? Maybe along lines of covariate/baseline and/or continuous/binary?

JanionNevill, line 10: @tommorris168 Can you check the contents of this function please?

R/create_covariate_summary_plot.R
```diff
@@ -0,0 +1,158 @@
...
+#' Paste the caption text
+#'
+#' @param plot_type Text string to describe type of plot. Can be "baseline risk" or "covariate"
+#' @param error_bar_text Text string to explain the error bar (optional)
+#' @return Text string to be used for caption
+
+PasteCaptionText <- function(caption_setting, error_bar_text = NULL) {
+  
+  # Short lines because the line lengths are not reactive to plot width and I haven't found a fix
+  caption_text <- paste('The plotted', caption_setting, 'value is the same for all treatment arms across a study
```
JanionNevill, line 149: Some of this text is being dropped of the edge of the plot area because the line is too long, or the plot is too short. Also there does seem to be a lot of dead space in the page.
image
Since we've now dropped the requirement for a sensitivity analysis, there is a lot more real estate on offer.

JanionNevill (changes requested):
A few things to tidy up, but looks to be working

R/create_covariate_summary_plot.R
JanionNevill: I've just seen that this is a known issue, so Is this issue explicit in another task?

R/SummaryPanel4a.R
JanionNevill: The momentary error is down to this being rendered later, so the value is not available when the plot is made.

The best fix would be to wrap the `shinyWidgets::updateRadioGroupButtons()` in an `observe` with a priority of 1 to make sure that the input is available before the plot is rendered.

As an aside, it's best to give all inputs and outputs unique names to avoid confusion.

global.R
```diff
@@ -1,5 +1,10 @@
 ###### MetaInsight ######
 
+# The version of BUGSnet on the CRSU GitHub account needs to be used 
+# remotes::install_github("https://github.com/CRSU-Apps/BUGSnet", ref="data.plot_edits")
+# You may wish to remove this version of BUGSnet afterwards using
+# remove.packages("BUGSnet")
```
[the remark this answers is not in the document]
nabury: Created a general task to check the wiki at the end of the project

tommorris168 (changes requested):
I thought there must be a mistake in the binary case because the error bars were going below 0. You can see it in the default dataset. It turns out they weren't, it was just the axes were being labelled with 0 because any value below 0.005 was being rounded to 0. I think the solution is a dynamic number of digits in the round function here: `scale_y_continuous(labels = function(x) round(plogis(x), digits = 2)).`

In the text underneath I would prefer this:
The plotted baseline risk value is the same for all treatment arms across a study and is the outcome in the reference treatment arm.

In the binary case I prefer this text for the error bars:
Error bars: logit(outcome) +/- 1.96 * se(logit(outcome))

In both continuous and binary we can delete "it is assumed that the data are normally distributed", because anyone interested will be able to tell from the +/- 1.96.

I don't suppose the study names on the x-axis can be better aligned? They're too far to the right.

R/SummaryPanel4a.R
nabury: There's a fatal error without it, because of the width = ... argument.

JanionNevill: Ah ok, that's a shame. Can you add a comment explaining this please?

R/SummaryPanel4a.R
nabury: > The momentary error is down to this being rendered later, so the value is not available when the plot is made.
> 
> The best fix would be to wrap the `shinyWidgets::updateRadioGroupButtons()` in an `observe` with a priority of 1 to make sure that the input is available before the plot is rendered.

I tried this and am still getting the error.

nabury: > As an aside, it's best to give all inputs and outputs unique names to avoid confusion.

Changed

R/create_covariate_summary_plot.R
nabury: Taken out the plot settings and data mutation into separate functions, which makes the remainder more readable

nabury:
> In the text underneath I would prefer this: The plotted baseline risk value is the same for all treatment arms across a study and is the outcome in the reference treatment arm.
> 
> In the binary case I prefer this text for the error bars: Error bars: logit(outcome) +/- 1.96 * se(logit(outcome))
> 
> In both continuous and binary we can delete "it is assumed that the data are normally distributed", because anyone interested will be able to tell from the +/- 1.96.
> 

Changes made as suggested

> I don't suppose the study names on the x-axis can be better aligned? They're too far to the right.

Will look at this in the separate BUGSnet task

R/create_covariate_summary_plot.R
nabury: I will investigate this further when the plot is rotated as a workaround now may break after that. (Edited the BUGSnet Jira task)

R/create_covariate_summary_plot.R
JanionNevill: Can you update the function names please? Because R has a global namespace, simple general function names are too likely to clash with other functions, either our own, or those in a package.

Something like `CreateCovariateSummaryPlotSettings` and `MutateCovariateSummaryData` would be safer.

I wish we could have simpler names, but this is safer.

R/SummaryPanel4a.R
JanionNevill: I managed to fix the error by swapping the `input$toggle_covariate_baseline` to `ifelse(is.null(input$toggle_covariate_baseline), 'Baseline risk', input$toggle_covariate_baseline)` in the call to `CreateCovariateSummaryPlot()`.

This is also a bit cleaner as the check in the plot function: `covariate_or_baseline == "Baseline risk"` evaluates to an empty vector `logical(0)` when `covariate_or_baseline` is `NULL`.

R/create_covariate_summary_plot.R
```diff
@@ -0,0 +1,213 @@
...
+    
+    if (metaoutcome == "Continuous") {
+        
+      # Error bar text for continuous outcomes
+      error_bar_text <- "Error bars: mean +/- 1.96 * SD / sqrt(N)"
+      
+    } else if (metaoutcome == "Binary") {
+      
+      # Error bar text for binary outcomes
+      error_bar_text <- "logit(outcome) +/- 1.96 * se(logit(outcome))"
```
tommorris168, line 29: "Error bars: " is missing from the beginning here.

R/SummaryPanel4a.R
nabury: > This is also a bit cleaner as the check in the plot function: `covariate_or_baseline == "Baseline risk"` evaluates to an empty vector `logical(0)` when `covariate_or_baseline` is `NULL`.

I never liked that line so this is much better.

JanionNevill (approved):
Looks good to me. Tom still has some comments to approve

tommorris168:
All my requested changes look good, but the download doesn't work for me.

nabury:
> All my requested changes look good, but the download doesn't work for me.

Good spot. Should be fixed now.

Note the change in path for BUGSnet when running locally

tommorris168 (changes requested):
When I loaded a dataset with a covariate, the text underneath the covariate graph was wrong, because it said the value displayed was the outcome in the reference arm, which is the baseline risk text. Also the text underneath the baseline risk graph was going off the page to the left, although it was fine in the download. The dataset I loaded was "Cont_long_continuous_cov".

tommorris168 (changes requested):
We don't seem to be on the same page. For the baseline risk model, the covariate is the outcome in the reference arm, which is stated on the screen. And since some studies may not have a reference arm, there is no value displayed for those studies. So the text is correct for baseline risk.
For the ordinary meta-regression model, the covariate is whatever has been uploaded. So the text that says it's the outcome in the reference arm is wrong. Moreover, the covariate values are being (correctly) displayed for all studies, regardless of which studies have a reference arm. So the text that says they are only displayed for studies that have the reference arm is also wrong.

tommorris168 (approved):
Yes

PR #127, opened 2024-01-15
MI-39 Added app name to analytics cookie module

JanionNevill:
Added app name to cookie module


tommorris168 (approved):
Looks fine, no comments.

PR #129, opened 2024-01-16
MI-40 Data column case insensitivity

JanionNevill:
Added letter case fixing in to data wrangling. After this is merged, I will merge into `feature-meta-regression` and add tests for covariate column.

### DIFF
--- a/R/DataWrangling.R
+++ b/R/DataWrangling.R
@@ -12,21 +12,26 @@ CleanData <- function(data) {
 #' @param data Data frame in which to search for treatment names
 #' @return Vector of all treatment names
 FindAllTreatments <- function(data) {
-  if ('T' %in% colnames(data)) {
-    # Long format
-    return(unique(data$T))
-  } else {
-    # Wide format
-    all_treatments <- c()
-    for (col in paste0('T.', seq(6))) {
-      if (col %in% colnames(data)) {
-        all_treatments <- c(all_treatments, data[[col]])
-      } else {
-        break
+  # Regular expression explanation:
+  # ^ = Start of string
+  # (?i) = Ignore case for matching
+  # (\\.[0-9]+)? = Optional group of full stop, followed by at least one digit
+  # $ = End of string
+  treatment_column_matches <- stringr::str_match(names(data), "^(?i)T(\\.[0-9]+)?$")
+  treatment_column_names <- treatment_column_matches[!is.na(treatment_column_matches)]
+  
+  treatment_names = unlist(
+    sapply(
+      treatment_column_names,
+      function (nom) {
+        treatments <- data[[nom]]
+        return(treatments[!is.na(treatments)])
       }
-    }
-    return(unique(all_treatments[!is.na(all_treatments)]))
-  }
+    )
+  )
+  
+  # Wrapped in c() to convert to an unnamed vector
+  return(unique(c(treatment_names)))
 }
 
 #' Create a copy of a vector with the given item as the first element.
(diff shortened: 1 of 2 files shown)


R/DataWrangling.R
```diff
@@ -12,21 +12,26 @@ CleanData <- function(data) {
...
+  # ^ = Start of string
+  # (?i) = Ignore case for matching
+  # (\\.[0-9]+)? = Optional group of full stop, followed by at least one digit
+  # $ = End of string
+  treatment_column_matches <- stringr::str_match(names(data), "^(?i)T(\\.[0-9]+)?$")
+  treatment_column_names <- treatment_column_matches[!is.na(treatment_column_matches)]
+  
+  treatment_names = unlist(
+    sapply(
+      treatment_column_names,
```
tommorris168, line 25: I don't think you need unlist here, because sapply returns a vector. There are two other instances of this.

tests/testthat/test-DataWrangling.R
```diff
@@ -107,6 +107,70 @@ test_that("FindExpectedReferenceTreatment() returns first matching treatment whe
   expect_equal(!!result, "Placebo")
 })
 
+test_that(".FixColumnNameCases() fixes cases for continuous long data", {
+  data <- CleanData(read.csv("Cont_long.csv"))
+  names(data) <- c("sTuDy", "t", "n", "mEaN", "sD")
+  allowed_names = c("Study", "T", "N", "Mean", "SD")
+  
+  expect(all(!names(data) %in% allowed_names), failure_message = "Column names were not setup for the test correctly.")
+
```
tommorris168, line 116: The ! should be in front of all( ). The same problem in several other places

tests/testthat/test-DataWrangling.R
```diff
@@ -107,6 +107,70 @@ test_that("FindExpectedReferenceTreatment() returns first matching treatment whe
...
+  
+  wrangled_data <- .FixColumnNameCases(data, "Continuous")
+  
+  expect_equal(colnames(wrangled_data), allowed_names,
+               label = format_vector_to_string(colnames(wrangled_data)),
+               expected.label = format_vector_to_string(allowed_names))
+})
+
+test_that(".FixColumnNameCases() fixes cases for continuous wide data", {
+  data <- CleanData(read.csv("Cont_wide.csv"))
```
tommorris168, line 125: This test_that fails.

tommorris168 (changes requested):
A failed test_that plus some minor issues.

tests/testthat/test-DataWrangling.R
JanionNevill: I think it is correct as it is. I am asserting that all of the names are not in `allowed_names`, not that at least one isn't.

tests/testthat/test-DataWrangling.R
JanionNevill: Can you show me the error please? I have all tests passing locally, with no outgoing changes.

R/DataWrangling.R
JanionNevill: Unfortunately removing the `unlist()` call causes errors. This seems to be because it's not guaranteed to be a vector. The return can be a matrix or list, depending on the output from the applied function.

tests/testthat/test-DataWrangling.R
tommorris168: I can see where you're coming from.

tests/testthat/test-DataWrangling.R
tommorris168: If I run test_that.R everything passes, but if I run the test_thats individually, this one fails:

── Failure: .FixColumnNameCases() fixes cases for continuous wide data ─────────
c("Study", "T.1", "N.1", "Mean.1", "SD.1", "T.2", "N.2", "Mean.2", "SD.2", "T.3", "N.3", "Mean.3", "SD.3", "NA", "NA", "NA", "NA") not equal to c("Study", "T.1", "N.1", "Mean.1", "SD.1", "T.2", "N.2", "Mean.2", "SD.2", "T.3", "N.3", "Mean.3", "SD.3").
Lengths differ: 17 is not 13

Error: Test failed

PR #135, opened 2024-02-07
MI-43 ranking naming

JanionNevill:
Fixed minor bug with naming. Also did minor code tidy-up.


tommorris168 (approved):
All fine. I couldn't reproduce the issue at first because it works in Open Office, but when I uploaded to Excel online the problem arose.

PR #136, opened 2024-02-08
MI-44 move outcome selection

JanionNevill:
Moved outcome selection and removed banners.


tommorris168 (changes requested):
Much nicer...but please change the message to refer to the "Load Data tab" rather than the "data upload page".

tommorris168 (changes requested):
One more change: the error message on line 133 in DataInputPanel.R is now out of date.

tommorris168 (approved):
Yes

PR #137, opened 2024-02-09
Mr 39 baseline risk tabs 3 and 4

tommorris168:


R/bayesian/ranking/ranking_forest_panel.R
```diff
@@ -80,10 +80,65 @@ ranking_forest_panel_server <- function(
...
+    filename_prefix,
+    title_prefix
+) {
+  moduleServer(id, function(input, output, session) {
+    
+    # Forest plots for ranking panel (different style due to using 'boxes' in UI) CRN
+    # All studies #
+    output$gemtc2 <- renderPlot({
+      png("forest.png")  # initialise image
+        bnma::network.forest.plot(model(), only.reference.treatment = TRUE)
```
JanionNevill, line 117: Unnecessary indent here

R/bnma_analysis.R
```diff
@@ -124,3 +124,97 @@ BaselineRiskDicTable <- function(br_model){
...
+    for(col in 1:dim_median){
+      lower_ci[row, col] <- round(
+        as.numeric(
+          substr(x = median_ci_table[row, col],
+                 start = 2,
+                 stop = as.vector(
+                   gregexpr(pattern = ",",
+                            text = median_ci_table[row, col])[[1]]
+                   )[1] - 1
+                 )
```
JanionNevill, line 155: This chunk to extract the first number in the string can be simplified to:
```
median_br[row, col] <- median_ci_table[row, col] %>%
  stringr::str_extract(pattern = "([-0-9\\.]+)") %>%
  as.numeric() %>%
  round(digits = 2)
```
(My best attempt to make it tidyverse-style)

I think that's more readable than using `gregexpr` to get indices, though that does work.

R/bnma_analysis.R
```diff
@@ -124,3 +124,97 @@ BaselineRiskDicTable <- function(br_model){
...
+                 start = as.vector(
+                   gregexpr(pattern = ",",
+                            text = median_ci_table[row, col])[[1]]
+                 )[1] + 1,
+                 stop = as.vector(
+                   gregexpr(pattern = ",",
+                            text = median_ci_table[row, col])[[1]]
+                 )[2] - 1
+          )
+        ), digits = 2
```
JanionNevill, line 176: Same tidy-up as above possible.

R/bnma_analysis.R
```diff
@@ -124,3 +124,97 @@ BaselineRiskDicTable <- function(br_model){
...
+      upper_ci[row, col] <- round(
+        as.numeric(
+          substr(x = median_ci_table[row, col],
+                 start = as.vector(
+                   gregexpr(pattern = ",",
+                            text = median_ci_table[row, col])[[1]]
+                 )[2] + 1,
+                 stop = nchar(median_ci_table[row, col]) - 1
+          )
+        ), digits = 2
```
JanionNevill, line 193: As above

tests/testthat/test-bnma_analysis.R
```diff
@@ -209,3 +209,41 @@ test_that("BaselineRiskRegression() sets RNGs correctly and gives reproducible o
...
   
 })
+
+
+
+test_that("BaselineRiskRelativeEffectsTable() has the correct format", {
+  
+  rel_eff_table <- matrix(c(NA, "[-0.1,0,0.1]", "[-2.3,-1,4]",
+                            "[3,4,5]", NA, "[2,6,8.3]",
+                            "[-7,4,11]", "[15,16,17]", NA), nrow = 3)
```
JanionNevill, line 219: Is it worth adding some numbers with more than 2 decimal places to ensure that it's rounding as expected?

JanionNevill (approved):
Minor changes

JanionNevill (changes requested):
Changes in other review...

R/bnma_analysis.R
tommorris168: I have used str_extract_all as I agree it's more readable. I've also put the three extractions in the same for-loop. But I haven't used the pipe operator because I don't like it.

R/meta_regression/treatment_comparison_page.R
```diff
@@ -49,6 +49,12 @@ covariate_treatment_comparisons_page_server <- function(
     output$package <- reactive({"gemtc"})
     outputOptions(x = output, name = "package", suspendWhenHidden = FALSE)
     
+    output$nodesplit_text <- renderText({
+      "This table only contains the estimates from the network meta analysis,
+        i.e. does not contain estimates from pairwise meta-analysis which only contains direct evidence.
+        If you would like to obtain the pairwise meta-analysis results, please run 4c-4. Nodesplit model"
+    })
```
JanionNevill, line 56: Since this text is not actually reactive, it can be put into the conditional panel directly, wrapping it in a `p()` tag

R/bnma_analysis.R
JanionNevill: Great idea to combine the loops. I hadn't noticed that.
One extra improvement can therefore be to extract the numbers once, then assign them, rather than running the same regex 3 times:
```
interval <- round(
    as.numeric(
        stringr::str_extract_all(
            string = median_ci_table[row, col],
            pattern = "[-0-9\\.]+"
        )[[1]]
    ), digits = 2
)
lower_ci[row, col] <- interval[1]
median_br[row, col] <- interval[2]
upper_ci[row, col] <- interval[3]
```

JanionNevill (changes requested):
Couple of tiny final fixes. Otherwise all looks good

tommorris168:
After making the commit I tried to change the name from "Response to review" to "Response to second review", so I've ended up with three new commits, two of which are identical and one that does nothing. Just look at "Response to second review".

JanionNevill (approved):
All good

PR #139, opened 2024-02-13
MI-45 consent storage

JanionNevill:
Updated from cookies to browser local storage.

I couldn't think how to test it because the code relies on having a script in the UI, so a server test won't work. Any suggestions are welcome.


www/storage.js
```diff
@@ -0,0 +1,69 @@
+// Based on code found at:
+// https://github.com/RinteRface/shinyLocalStorage/tree/main
```
RyanJField, line 2: Please include the MIT licence requirements:
https://github.com/RinteRface/shinyLocalStorage/tree/main?tab=MIT-1-ov-file#readme

```
Copyright (c) 2024 David Granjon

Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:

The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
```

R/local_storage.R
```diff
@@ -0,0 +1,92 @@
...
+#' 
+LocalStorage <- R6::R6Class(
+  classname = "Local Storage",
+  public = list(
+    #' @description
+    #' Create a new instance of this [R6][R6::R6Class] class.
+    initialize = function() {
+      # Get root session so that storage is not tied to module namespace
+      private$session <- shiny::getDefaultReactiveDomain()$rootScope()
+      private$storage <- reactive({
```
RyanJField, line 15: You've scoped the `getDefaultReactiveDomain`, but not this `reactive`, might need some consistency.

RyanJField (changes requested):
This looks good, only needs the MIT requirements on the storage.js file.

I tested it in an iFrame and it appeared to work, but please test on shinyapps with an iFrame.

R/local_storage.R
JanionNevill: Fixed

www/storage.js
JanionNevill: Added

JanionNevill:
> please test on shinyapps with an iFrame.

I tested it in an iframe pointing to shinyapps.io in a dummy app running locally in the browser. Seemed to work fine, but didn't share storage between the embedded instance and the direct access instance.

RyanJField (approved):

PR #151, opened 2024-03-15
mr-52-fix-contributions-in-contribution-matrix

tommorris168:
Weinblatt1999 now has a contribution when `weight_or_contribution = "contribution"`.

If you look at the row for Strand2006 and compare `weight_or_contribution = "contribution"` to `weight_or_contribution = "weight"` you see that this study now has zeros when `weight_or_contribution = "contribution"`. In this case it is correct, because both arms have the same outcome (R = 5, N = 40), so the treatment effect is zero and the arms cancel each other out.


JanionNevill (commented):
I'm surprised that this wasn't caught by a test. Can a test be written to cover this please?

tommorris168:
Closing, and making the fix on a new branch where tests will be added as well.

JanionNevill:
@tommorris168 Can this branch be deleted?

PR #153, opened 2024-03-21
Mr 53 add contribution matrix tests

tommorris168:


R/contribution_matrix.R
```diff
@@ -724,8 +724,12 @@ CreateContributionMatrix <- function(data, treatment_ids, outcome_type, outcome_
                                                treatments = treatments,
                                                outcome_type = outcome_type,
                                                outcome_measure = outcome_measure)$effect_sizes$Effect
-    #Multiply the n-th row of 'contribution' by the n-th element of 'effect_sizes'
-    contribution <- contribution * effect_sizes
```
BalancedmonkeY, line 728: I don't understand why this doesn't do the job?
When I did the following:
```
matrix <- matrix(1:6, nrow = 2, ncol = 3)
test_effects <- c(2,3)
matrix * test_effects
```
This did indeed multiply the elements in the first row by 2, and the elements in the second row by 3.
Am I missing something?

tommorris168: Sorry, I got myself confused. Each _column_ should be multiplied by a single treatment effect, not each row as I have written on Jira. I will change the text on Jira. The confusion is because of the transposing (again). It would be each row if the multiplication took place after transposing.

tests/testthat/test-contribution_matrix.R
```diff
@@ -374,3 +374,350 @@ test_that("CreateContributionMatrix() produces a matrix of the correct format wh
...
+                                                  outcome_type = "Continuous",                           
+                                                  outcome_measure = "MD",
+                                                  effects_type = "fixed",
+                                                  cov_parameters = "shared",
+                                                  basic_or_all_parameters = "all",
+                                                  study_or_comparison_level = "comparison",
+                                                  absolute_or_percentage = "absolute",
+                                                  weight_or_contribution = "weight")
+  
+  basic_contribution_matrix <- CreateContributionMatrix(data = data,
```
BalancedmonkeY, line 585: `basic_contribution_matrix` isn't used anywhere in the test - remove?

R/contribution_matrix.R
```diff
@@ -724,8 +724,12 @@ CreateContributionMatrix <- function(data, treatment_ids, outcome_type, outcome_
...
+    #Create an effect size matrix in order to do the multiplication in the next step
+    effect_size_matrix <- matrix(rep(effect_sizes, each = nrow(contribution)),
+                                 nrow = nrow(contribution)
+                                 )
+    #Multiply the n-th column of 'contribution' by the n-th element of 'effect_sizes'
+    contribution <- contribution * effect_size_matrix
```
BalancedmonkeY, line 732: Whilst your method does work and is only one extra line, I think you can keep it to 1 line rather than 2 with
```
contribution <- t(t(contribution)*effect_sizes)
```

BalancedmonkeY (changes requested):

tests/testthat/test-contribution_matrix.R
tommorris168: Correct, I had wanted to use it but there was a reason why I couldn't so had to do it manually. I think it was because it was impossible to tell which values had been positive and which negative before taking the absolute values.

R/contribution_matrix.R
tommorris168: I tried transposing `effect_sizes` but it didn't work. I did not think about transposing `contribution`!

tests/testthat/test-contribution_matrix.R
```diff
@@ -374,3 +374,350 @@ test_that("CreateContributionMatrix() produces a matrix of the correct format wh
...
+                                                  basic_or_all_parameters = "all",
+                                                  study_or_comparison_level = "comparison",
+                                                  absolute_or_percentage = "percentage",
+                                                  weight_or_contribution = "weight",
+                                                  full_output = FALSE)
+  
+  expected_contribution_matrix <- read.csv("Donegan_contribution_matrix_expected.csv")
+  expected_contribution_matrix <- as.matrix(expected_contribution_matrix[, 2:7])
+  colnames(expected_contribution_matrix) <- c("1:2_d", "1:3_d", "2:3_d", "1:2_beta", "1:3_beta", "2:3_beta")
+  rownames(expected_contribution_matrix) <- rownames(contribution_matrix)
```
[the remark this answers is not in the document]
BalancedmonkeY: I had initially envisioned it on every test, but that is probably overkill, and all the tests where this is the case are together and follow the same pattern, so it's should be understandable :)

R/contribution_matrix.R
BalancedmonkeY: Haha, I mostly thought of it from the conversation we had about which way round the multiplication should be, and you said it's because of the transposition :D

BalancedmonkeY (approved):

tommorris168:
I was going to merge it after MR-43.

PR #155, opened 2024-04-05
MI-51 validate data types

JanionNevill:
Reformatted data definitions and added validation for data typing. This could be updated in future to have the data definitions as R6 objects, but that wasn't necessary for this change.


RyanJField (commented):
This looks good, the only thing is I can upload an empty xlsx file and it will crash the app.

JanionNevill:
> This looks good, the only thing is I can upload an empty xlsx file and it will crash the app.

I've added handling for empty files and files containing headers only

RyanJField (approved):
Looks good.

PR #157, opened 2024-04-16
Troubleshooting patch

BalancedmonkeY:
Also added new home page image and text to say we can now upload .xlsx.

To check the troubleshooting link is working properly, you'll probably need to run the app within a browser.


RyanJField (approved):

JanionNevill (approved):
Release actions are all done 👍

PR #158, opened 2024-05-03
Mr 58 novel plot baseline risk

tommorris168:


R/meta_regression/covariate_value_panel.R
```diff
@@ -87,13 +87,21 @@ covariate_value_panel_server <- function(id, covariate_type, covariate_data) {
       return(max(covariate_data()))
     })
     
+    # Mean covariate value in the data
+    mean_value <- reactive({
+      if (is.null(covariate_data())) {
+        return(NULL)
```
JanionNevill, line 93: Again, I'm a little confused by the necessity of this check.

tests/testthat/test-contribution_matrix.R
```diff
@@ -641,7 +641,10 @@ test_that("CalculateContributions() gathers indirect covariate effect contributi
 
 test_that("CreateContributionMatrix() produces correct output for the Donegan example", {
   #Load data
+  donegan_original <- read.csv("Donegan_contribution_matrix_data.csv")
   donegan <- read.csv("Donegan_for_MetaInsight.csv")
+  donegan_ids <- data.frame(Number = 1:3, Label = as.character(1:3))
+  donegan_studies <- unique(donegan$Study)
```
JanionNevill, line 647: These additional lines don't seem to be used. What purpose do they serve?

R/meta_regression/treatment_comparison_page.R
```diff
@@ -97,7 +97,9 @@ treatment_comparisons_page_baseline_risk_server <- function(
     output$downloadbaye_comparison <- downloadHandler(
       filename = 'baseline_risk_comparison.csv',
       content = function(file) {
-        write.csv(baye_comp(model(), outcome_measure()), file)
+        write.csv(
+          BaselineRiskRelativeEffectsTable(bnma::relative.effects.table(model(), summary_stat = "ci")), file
```
JanionNevill, line 101: Can you explain this change please? It looks like the function `BaselineRiskRelativeEffectsTable()` is now going to be used for both covariate analysis and baseline risk analysis.

JanionNevill (changes requested):
There are a couple of things that I need clarification on, otherwise, it looks good.

R/contribution_matrix.R
```diff
@@ -813,6 +817,9 @@ CreateContributionMatrix <- function(data, treatment_ids, outcome_type, outcome_
   treatments <- treatment_ids$Label
   #Unduplicated covariate values (one per study)
   covariate <- unique(dplyr::select(data, starts_with(c("Study", "covar."))))$covar.
+  if (any(is.na(covariate))) {
+    stop("Missing covariate values are not allowed")
```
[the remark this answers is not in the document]
tommorris168: There were missing values early on in the task before I updated GetReferenceOutcome( ) to get the imputed values. So I think you're right that this isn't necessary anymore. I'd like to leave it in case the function is used outside MetaInsight.
I believe it just showed the red message when there were missing values. Actually I had to remove the tryCatch statement when doing the task because I couldn't see the error message when anything went wrong.

R/meta_regression/covariate_value_panel.R
tommorris168: Same as above, but this time I'm happy to remove it if you want?

tests/testthat/test-contribution_matrix.R
tommorris168: If I comment out either line the test fails. `donegan_ids` is a parameter in `CreateContributionMatrix( )` and `donegan_studies` is used in overwriting `CreateVMatrix( )`.
It's strange because the test passed in the original task. I can only think it's because these were both in my environment.

R/meta_regression/treatment_comparison_page.R
tommorris168: You can see why I made the mistake in the first place. There are two server functions!

R/meta_regression/covariate_value_panel.R
JanionNevill: Since this is in a reactive covered by the other check, I think this one should be removed.

JanionNevill (changes requested):
Just one minor removal, then we're good.

R/meta_regression/covariate_value_panel.R
tommorris168: Are you sure? I added `mean_value`, but `min_value` and `max_value` were already there and they have the check too.

JanionNevill: Oh, I see... Let's just leave it, then. It doesn't hurt to have anyway.

JanionNevill (approved):
👍

PR #160, opened 2024-05-07
Selected studies excluded

tommorris168:


BalancedmonkeY (changes requested):
Just for extra consistency - can we also change tab 3g-4 to say Deviance data for selected studies excluded.

Also, possibly change the title on tab 3f that currently says "Deviance report for all studies and the sensitivity analysis". I.e. don't have the phrase 'sensitivity analysis' as it's not used anywhere else, so want to keep things as clear and simple as possible.

BalancedmonkeY (approved):

PR #162, opened 2024-05-20
MR-55 confidence regions

JanionNevill:
Added background calculation for confidence regions.


JanionNevill:
Made PR into wrong branch

PR #163, opened 2024-05-20
MR-55 confidence regions

JanionNevill:
Added background task to calculate confidence regions.


JanionNevill:
> I've found two problems. In the certolizumab data set (attached)
> 
> 1. one of the colours doesn't match between the line and the region.
> 2. if you add all treatments, add all the confidence regions, then remove them one by one, at some point it fails.

Those 2 issues turned out to have the same root cause. I fixed this by adding `NA`s for intervals where a region was present, and vice versa.

tommorris168 (approved):
Getting close to completion.

PR #168, opened 2024-05-23
MR-54 Regression Download Button

JanionNevill:
Added download button to regression plot


tommorris168 (changes requested):
Have you tried the png download? It has come out really funky for me. It may need some attention...

tommorris168 (approved):
Approved!

PR #169, opened 2024-05-25
Mr 61 baseline risk confidence regions

tommorris168:
`CalculateConfidenceRegionsBnma()` and `.FindConfidenceIntervalBnma()` both work, (as does `BnmaRelativeEffects()`, though that is not for review in this task, it's in mr-60). If I remove all the `ExtendedTask` code then it all works.

I have tried to modify the `ExtendedTask$new` code to call `CalculateConfidenceRegions()` or `CalculateConfidenceRegionsBnma()` depending on the value of `package`, to no avail. I have commited three attempts which are commented out. I have also tried adding `package` as another parameter to the function within `ExtendedTask$new` but that didn't work either. I'm stuck now and would appreciate some help!

### DIFF
--- a/R/bnma_analysis.R
+++ b/R/bnma_analysis.R
@@ -138,7 +138,7 @@ BaselineRiskRegression <- function(br_data, treatment_ids, outcome_type, ref,  e
 #'  covariate_min = Vector of minimum covariate values directly contributing to the regression.
 #'  covariate_max = Vector of maximum covariate values directly contributing to the regression.
 BaselineRiskModelOutput <- function(data, treatment_ids, model, outcome_measure) {
-  
+
   treatments <- model$network$Treat.order
   reference <- unname(treatments[1])
   comparator_names <- unname(treatments[-1])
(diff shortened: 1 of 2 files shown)


R/bnma_analysis.R
```diff
@@ -267,6 +267,90 @@ GetReferenceOutcome <- function(data, treatment_ids, outcome_type, observed, mod
...
+CalculateConfidenceRegionsBnma <- function(model_output) {
+  
+  mtc_results <- model_output$mtcResults
+  treatments <- mtc_results$network$Treat.order
+  
+  confidence_regions <- list()
+  confidence_intervals <- list()
+  
+  for (treatment_name in model_output$comparator_names) {
+    parameter_name <- glue::glue("d[", which(treatment_name == unname(treatments)), "]")
```
JanionNevill, line 293: I think that's not the intended way to use glue. The usual way would be:
`glue::glue("d[{which(treatment_name == unname(treatments))}]")`

Optional if you want to change it.

JanionNevill (approved):
One minor optional change

PR #170, opened 2024-06-04
Mr 34 rotate plot

nabury:
Meta-regression summary plot rotation and associated fixes


tommorris168 (approved):
Approved with a tweak to the height of the download.

PR #173, opened 2024-06-06
MR-30 move novel plot

JanionNevill:
Also added spinner when model is running

### DIFF
--- a/R/meta_regression/regression_plot_panel.R
+++ b/R/meta_regression/regression_plot_panel.R
@@ -203,7 +203,19 @@ regression_plot_panel_ui <- function(id) {
           selectize = FALSE
         )
       )
-    )
+    ),
+    
+    h4(style = "text-align:left",
+      "This graph was adapted from",
+      tags$em("Graphs of study contributions and covariate distributions for network meta-regression"),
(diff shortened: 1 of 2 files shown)


R/meta_regression/regression_plot_panel.R
```diff
@@ -211,6 +221,12 @@ regression_plot_panel_ui <- function(id) {
 regression_plot_panel_server <- function(id, data, covariate_title, covariate_name, model_output, treatment_df, outcome_type, outcome_measure, reference, package = "gemtc") {
   shiny::moduleServer(id, function(input, output, session) {
     
+    # This will show a spinner while the model calculates, but will not render anything once it has completed
+    output$spinner <- renderUI({
+      model_output()
+      NULL
```
tommorris168, line 239: Out of interest, what does NULL do here?

tommorris168 (approved):
Approved. I have snuck a citation in.

R/meta_regression/regression_plot_panel.R
JanionNevill: It returns null to the render UI so that it renders nothing. I tried to add a spinner without doing this, but I couldn't get the events to trigger in the correct order.

PR #176, opened 2024-06-18
mr-65-na-covariate-values

tommorris168:
The issue occurs whenever the reference treatment is changed manually on the load data tab.

### DIFF
--- a/R/DataWrangling.R
+++ b/R/DataWrangling.R
@@ -320,6 +320,14 @@ ReorderColumns <- function(data, outcome_type) {
   return(data[, reordering_indices])
 }
 
+#' Sort the data by Study then T
+#' 
+#' @param data Data frame to sort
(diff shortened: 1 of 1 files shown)


JanionNevill:
I'm a little confused. Can you explain to me the steps to recreate the original issue please?

tommorris168:
Sorry, it's not whenever the reference treatment is changed manually, it's whenever the reference treatment is not the first one to feature in the dataset. If you load the attached data you will see that placebo is selected as the default reference, but it isn't first in the list and it isn't the first one in the data. Then run the model and look at the novel plot. All the covariate values are NA. Sorting the data as in the commit fixes the problem.
[Non_opioids_long.csv](https://github.com/user-attachments/files/15902571/Non_opioids_long.csv)

JanionNevill (approved):
✔

PR #177, opened 2024-06-19
mr-49-update-default-datasets

tommorris168:
I have removed 20 of the studies from the non-opioid dataset, which halves the running time (from ~23 seconds to ~11 seconds on my computer - it may still be too long in the web version).


BalancedmonkeY (changes requested):
See request about dataset text.

Also, when running tests, all of the tests for `DataInputPanel` seem to be failing, claiming they can't find the file. There was also one `bnma-analysis` test that failed.

BalancedmonkeY (approved):

PR #185, opened 2024-07-04
feature_meta_regression into dev

tommorris168:


JanionNevill (approved):
It's difficult to give this a meaningful review, but the feature branch is all working, and no extra commits have snuck in, so I'm happy.

There are a couple of minor changes needed, but these can be done on dev after this merge:
- In `regression_plot_panel_ui`, the text needs to be changed from "comparator" to "treatment" or maybe "comparison". There might be a better word which is widely used in the literature, but I don't know it.
- In `regression_plot_panel_server`, there needs to be an observer which clears `added_comparators` when the `treatment_df` reactive changes. This prevents strange errors when the data set is changed or when the reference treatment is changed.

PR #189, opened 2024-07-11
Hot fix for regression graph

JanionNevill:
Changed hidden character to quoted text for simplicity. The data had to be sorted alphabetically due to a limitation in {ggplot2}. This might be fixable in a better way, but that will take significant investigation. The committed change is a quick-fix for the issue at hand.


tommorris168 (approved):
Works offline.

JanionNevill:
@tommorris168 MY laptop is broken and I can no longer deploy anything. Can you redeploy please?

PR #196, opened 2024-08-19
MI-59 exclude all studies

JanionNevill:
All studies can now be excluded without crashing the app


JanionNevill:
Not meant to pull into main

PR #197, opened 2024-08-19
MI-59 exclude all studies

JanionNevill:
All studies can now be excluded without crashing the app


tommorris168:
You could have kept the first PR, because the branch to be merged into can be changed after creation. (In fact, for a while I didn't realise there was any other way of doing it!)

tommorris168 (commented):
This PR is fine, but there is a problem which I assume is from MI-28. If you check the boxes too quickly you go into an infinite loop, where they keep checking and unchecking.

JanionNevill:
> If you check the boxes too quickly you go into an infinite loop, where they keep checking and unchecking.

Nooooo! I thought I'd fixed that!

Can you give me a set of precise reproducible instructions please?

JanionNevill:
I've reverted the disconnected network issue on another branch. Are you happy for this to be merged now?

tommorris168:
It's crashing when I remove all the studies.

PR #199, opened 2024-08-27
Removed sensitivity analysis disconnected network detection

JanionNevill:
The disconnected network detection for the sensitivity analysis was bugging out by repeatedly selecting and deselecting items. This has been removed, but the initial disconnected data detection has been retained.

### DIFF
--- a/R/study_exclusions_panel.R
+++ b/R/study_exclusions_panel.R
@@ -83,11 +83,37 @@ study_exclusions_panel_server <- function(id, data, treatment_df, reference_trea
       studies <- isolate(all_studies())
       return(studies[!studies %in% main_connected_data()$Study])
     })
-
-    # Update "selections" for the initial data to be those excluded by taking the subnetwork
+    
+    # Update the checkboxes to tick the excluded studies
     observe({
-      selections(main_subnetwork_exclusions())
-    })
+      # Select check boxes for disconnected studies
+      disconnected_indices <- which(all_studies() %in% main_subnetwork_exclusions())
+      selected = all_studies()[disconnected_indices]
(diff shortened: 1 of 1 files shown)


R/study_exclusions_panel.R
```diff
@@ -83,11 +83,37 @@ study_exclusions_panel_server <- function(id, data, treatment_df, reference_trea
...
     observe({
-      selections(main_subnetwork_exclusions())
-    })
+      filtered_treatments <- FindAllTreatments(main_connected_data())
+      
+      lapply(
+        all_studies(),
+        function(study) {
+          index <- match(study, all_studies())
+          subElement <- glue::glue("#{session$ns('exclusionbox')} .checkbox:nth-child({index}) label")
```
RyanJField, line 103: I assume this was named previously as it's camel case not snake case?

R/study_exclusions_panel.R
```diff
@@ -83,11 +83,37 @@ study_exclusions_panel_server <- function(id, data, treatment_df, reference_trea
...
+          
+          study_treatments <- FindAllTreatments(data()[data()$Study == study, ])
+          
+          if (any(study_treatments %in% filtered_treatments)) {
+            # The 0ms delay is required to disable studies disconnected within the initial uploaded data
+            shinyjs::delay(0, shinyjs::enable(selector = subElement))
```
RyanJField, line 109: I assume there is not around the delay 0?

JanionNevill: It just doesn't work without it. I'm not entirely sure why, but it's not trivial to work out why.

R/study_exclusions_panel.R
JanionNevill: Fixed. Old habits die hard...

PR #201, opened 2024-09-10
MI-69 Fixed data upload issue

JanionNevill:
Added data_reset reactive val to trigger recalculation of sensitivity data when data is reset


nabury (approved):
Seems to be working as expected

PR #204, opened 2024-09-30
Mi 18 reinstall sfp v2

tommorris168:
There is a quirk in that the summary forest plot is responsive to the ranking radio buttons, whereas the main forest plot is not. For the default continuous data set, the two are aligned by default. For the old default binary dataset, they were also aligned by default, but for the new one they aren't (i.e. if OR is on the main forest plot then 1/OR is on the summary forest plot). The user has to change the ranking option to align them.

I loaded the new default binary data set into the old branch, and they were aligned. So the misalignment is probably not due to the data, but to other changes that have happened to the app since the old branch was opened. I don't think it's worth investigating so I have added some help text instead.

### DIFF
--- a/R/data_analysis_page.R
+++ b/R/data_analysis_page.R
@@ -211,6 +211,7 @@ data_analysis_page_server <- function(id, data, is_default_data, treatment_df, m
       freq_sub = freq_sub,
       bugsnetdt = bugsnetdt,
       bugsnetdt_sub = bugsnetdt_sub,
+      treatment_df = treatment_df,
       reference_alter = reference_alter
     )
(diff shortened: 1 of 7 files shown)


BalancedmonkeY:
Regarding the quirk, yes, the main forest plot has never been responsive to the ranking option. However, at Cochrane Colloquium last year I think somebody did suggest adding text on the x-axis indicating saying 'favour treatment' and 'favour reference'. This could be added as a task and for the plot to therefore be responsive to the ranking option.

BalancedmonkeY (approved):

PR #205, opened 2024-10-04
MI-70 Update nodesplit results

BalancedmonkeY:
Transformed nodesplit model output into a plot that takes into account the outcome type and ensure that comparisons shown in the forest plot match (i.e. reference treatment always 'second')

### DIFF
--- a/R/bayesian/nodesplit/nodesplit_panel.R
+++ b/R/bayesian/nodesplit/nodesplit_panel.R
@@ -80,7 +80,7 @@ nodesplit_panel_server <- function(
         if (input$download_format == "PDF") {
           pdf(file = file, width = 9)
         } else if (input$download_format == "PNG") {
-          png(file = file, width = 9)
+          png(file = file, width = 850)
         }
         plot(summary(model_nodesplit()), digits = 3)
         dev.off()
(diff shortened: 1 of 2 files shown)


R/bayesian/nodesplit/nodesplit_panel.R
```diff
@@ -47,19 +54,38 @@ nodesplit_panel_server <- function(
...
+        paste0("Nodesplit.",
+               input$download_format)
+      },
       content = function(file) {
-        write.csv(model_nodesplit(), file)
-      }
+        if (input$download_format == "PDF") {
+          pdf(file = file, width = 9)
+        } else if (input$download_format == "PNG") {
+          png(file = file, width = 9)
```
tommorris168, line 83: The png download doesn't work, which I believe is because the default units for the width is pixels, and 9 isn't big enough.

R/download_buttons.R
```diff
@@ -3,6 +3,20 @@
 
 ## Sizing functions for forest plots ##
 
+#' Calculate the pixel height of a node-split forest plot for a given number of comparisons
+#' 
+#' @param ncomp The number of comparisons in the plot
+#' @return The height of the plot in pixels
+NodePixels <- function(ncomp) {
```
tommorris168, line 10: Up to you, but I would have done this in one line:

return(max(400, ncomp * 80))

PR #206, opened 2024-10-18
Version 6.2.0

BalancedmonkeY:
MI-56 Wrangle study names
MI-18 Added summary forest plot and have it interacting with user options
MI-70 Back-transform and align orders of nodesplit model results


JanionNevill (approved):
👍

PR #208, opened 2024-10-31
Mi 68 regression plot binary covariate bug

tommorris168:
There was another bug. If the dataset contains a covariate that only takes the values 0 and 1, this is inferred to be a binary covariate. However, the user can select it to be continuous. In this case, the model didn't run because gemtc made the same inference, and it had not been overridden.

In hindsight we shouldn't have allowed the covariate type to be changed by the user, because the chances that a real dataset would contain a continuous covariate that only ever took the values 0 and 1 is negligible. I doubt anyone will ever use the option.

### DIFF
--- a/R/home_page.R
+++ b/R/home_page.R
@@ -47,7 +47,7 @@ home_page_ui <- function(id) {
             tags$li(
               "Meta-regression has been added. One covariate is allowed, which can be a new continuous or binary variable,
               or baseline risk. Two new graphs are available for meta-regression. The first displays the covariate values grouped by
-              treatment and study. The second plots the covariate against relative treatment effects, with confidence regions and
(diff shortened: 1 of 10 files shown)


JanionNevill (approved):
Everything seems to be working well.

I just had a thought that should the graph be hidden and replaced with "Please rerun model" when the covariate type or regression coefficient changes? That can go under another task though.

tommorris168:
> I just had a thought that should the graph be hidden and replaced with "Please rerun model" when the covariate type or regression coefficient changes? That can go under another task though.

Yes, that would be good. I will create a task.

At the moment the way this works for other options is inconsistent. For example, I just ran a model, then changed the effects from random to fixed, and nothing happened. Then I changed the outcome from MD to SMD, and it hid all the output.

PR #209, opened 2024-12-18
Fix bug

tommorris168:
The easiest way to see this has worked is to view the matrix itself rather than look at the regression plot. Put a browser() on line 908 of contribution_matrix.R and run `round(contribution_output, digits = 2)`. Compare the exchangeable random effects model on the continuous dataset with the shared or unrelated model. Notice the values in the top left of the matrix are similar. Do the same in dev and you'll see they are all zero in the exchangeable model, which is wrong.

### DIFF
--- a/R/meta_regression/contribution_matrix.R
+++ b/R/meta_regression/contribution_matrix.R
@@ -587,8 +587,8 @@ CheckSingularMatrix <- function(matrix){
   CheckSingularMatrix(t(X_star) %*% solve(V_star) %*% X_star)
   
   A <- solve(t(X_star) %*% solve(V_star) %*% X_star) %*% t(X_star) %*% solve(V_star)
-  A_row2_left <- A[(ncol(X) + 1):(ncol(X) + ncol(X_d)), 1:nrow(X)]
-  A_row3_left <- A[(ncol(X) + ncol(X_d) + 1):(ncol(X) + ncol(X_d) + ncol(X_beta)), 1:nrow(X)]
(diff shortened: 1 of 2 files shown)


BalancedmonkeY (changes requested):
I could not get it to show me the matrices as described in the pull request comment.

Regardless of this, I do think there should be a unit test for this calculation, to (1) check the fix works, and (2) the fact that this bug happened, illustrates the need for a test. 
Happy to discuss further on Teams, with potentially getting help with team members with more experience in unit testing.

tommorris168:
@BalancedmonkeY The problem all along with unit tests has been that we only have one example where we know what the matrix is supposed to be, and that's already in a unit test. What we could test in this case is that changing the options doesn't change the matrix "too much". Is that what you mean? To compare the contribution matrices under different assumptions and check the entries agree to within some (possibly arbitrary) limits?

tommorris168:
@BalancedmonkeY OK, that sounds like a good compromise. By the way, I think I know why I didn't spot this earlier. Most of my testing was done on the certolizumab dataset, which by chance has nrow(X) = ncol(X) in the exchangeable model. So correct output was being produced by accident! I will make sure the tests use a dataset without this property.

tommorris168:
There was no test for the random effects exchangeable model, so I've added that rather than checking for zeros in the top left. This test would have revealed the error in the first place.

Don't spend too long checking it as it's going to be changed again following the delta/theta corrections and my e-mail chat with Alex.

Ignore the failed tests that have been fixed already in another branch.

BalancedmonkeY (approved):
Thanks! I've checked that the new tests would have failed without the fix, and they now pass with the fix. Great job!

PR #210, opened 2025-01-01
Mi 57 contributions for nma

tommorris168:


R/meta_regression/contribution_matrix.R
```diff
@@ -223,6 +231,40 @@ CreateXMatrix <- function(data, studies, treatments, covar_centred, cov_paramete
     }
   }
   
+  #Populate the design matrix for NMA
+  if (cov_parameters == "none") {
```
BalancedmonkeY, line 235: Should really be an `else if`, as this will group them together illustrating that they are doing the same thing, and will also reduce computer checking if the first if statement is satisfied
(Also need to be done for line 195)

R/meta_regression/contribution_matrix.R
```diff
@@ -370,7 +421,7 @@ CheckSingularMatrix <- function(matrix){
 
 
 
-#' Intermediate function called within CreateContributionMatrix(), dealing with the fixed effects, unrelated or shared cases.
+#' Intermediate function called within CreateContributionMatrix(), dealing with the fixed effects unrelated or shared NMR, and NMA.
```
BalancedmonkeY, line 424: Might be clearer what is going on if the name `.WeightMatrixFixedUnrelatedShared` is also changed to reflect that this function is also usable for NMA. E.g. `.WeightMatrixFixedUnrelatedSharedNone`, or `.WeightMatrixFixedUnrelatedSharedNMA`?

R/meta_regression/contribution_matrix.R
```diff
@@ -467,7 +518,7 @@ CheckSingularMatrix <- function(matrix){
 
 
 
-#' Intermediate function called within CreateContributionMatrix(), dealing with the random effects, unrelated or shared cases.
+#' Intermediate function called within CreateContributionMatrix(), dealing with random effects unrelated or shared NMR, and NMA.
```
BalancedmonkeY, line 521: Same comment regarding the name

BalancedmonkeY (changes requested):
Specific comments are regarding formatting.

Main changes are that I think there are still some places that need the `cov_parameters` option of "none":

- `CalculateContributions` needs the option added
- Regarding `CreateContributionMatrix`, the return statement does not include the case of `cov_parameters` equalling "none"

tommorris168:
I did not originally plan to add `cov_parameters = "none"`  to `CalculateContributions` because the function is only used for the regression graph. However, it is possible it could be useful for CINeMA, and it was easy, so I've done it.

BalancedmonkeY (approved):

PR #212, opened 2025-01-07
MI-76 Updated to GA4

JanionNevill:
Copied from MetaDTA


tommorris168 (changes requested):
The modal isn't working. When I run the app I get "Error 'google_analytics.js' does not exist in current working directory". There is a file google_analytics2.js, which should presumably be replaced by google_analytics.js from MetaBayesDTA. I've tried copying the latter into the folder for this branch and the error message disappears, but no modal is shown.

If it was all working, I don't know how to check that offline. Would it have to go live before we could tell?

JanionNevill:
I've renamed the js file. I clearly hadn't tested it properly. I also renamed the cookie to `accept_analytics`

tommorris168 (approved):
The modal now shows, so I'm accepting it, but the caveat still stands that I don't know the analytics will actually work.

JanionNevill:
I've now confirmed that it is working

Issue #215, opened 2025-01-17
Deviance report residual deviance plot is slow to render


JanionNevill:
The problem is that it runs a new UME model every time it renders. This should be stored as an external `reactive` or `reactiveVal`, which can be evaluated and populated by the render function for future use. This should also take into account the model from the nodesplit page, as that is the same model, so could be reused.

PR #216, opened 2025-01-17
MI-78 hide bayesian

JanionNevill:
Everything is cleared when the model becomes invalid, and re-rendered when the model becomes valid.

This is somewhat intensive, but hiding the contents of `withSpinner()` results in infinite spinners, which is very misleading. I think that this is reasonable for now. The biggest problem is with the deviance report running a full UME model every time it renders. This is a bigger issue than this task, so I've raised it as issue #215 .

### DIFF
--- a/R/bayesian/invalid_model_panel.R
+++ b/R/bayesian/invalid_model_panel.R
@@ -18,16 +18,17 @@ invalid_model_panel_server <- function(id, model_valid) {
     id,
     function(input, output, session) {
       output$invalid_model <- renderUI({
-        if (model_valid()) {
-          return()
-        } else {
+        if (is.null(model_valid())) {
           return(
-            div(
-              h3("Please rerun model"),
-              style = ""
-            )
+            h3("Please run model")
+          )
+        } else if (!model_valid()) {
+          return(
+            h3("Please rerun model")
           )
         }
+        
+        return()
       })
     }
   )
--- a/R/bayesian/invalid_model_panel.R
+++ b/R/bayesian/invalid_model_panel.R
@@ -1,0 +1,34 @@
+
+#' Module UI for the invalid model panel.
+#' 
+#' @param id ID of the module.
+#' @return uiOutput for the panel.
+invalid_model_panel_ui <- function(id) {
+  ns <- NS(id)
+  return(uiOutput(outputId = ns("invalid_model")))
+}
+
+
+#' Module server for the nodesplit model panel.
+#' 
+#' @param id ID of the module
+#' @param model_valid Reactive containing whether the model is valid.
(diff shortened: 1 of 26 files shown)


R/bayesian/nodesplit/nodesplit_panel.R
```diff
@@ -43,36 +43,92 @@ nodesplit_panel_server <- function(
...
       {
-        return(
-          nodesplit(
-            data(),
-            treatment_df(),
-            metaoutcome(),
-            outcome_measure(),
-            model_effects()
-          )
+        nodesplit_model <- tryCatch(
```
nabury, line 46: Comments for new code

nabury (changes requested):
Functionality looks great - I have manually tested all paths I could think of and the behaviour is how I envisaged.

Tests ✅

One minor comment about the initial message displayed about re-running the model.

Other than that I would just like some narration on the new code to improve readability.

R/bayesian/invalid_model_panel.R
```diff
@@ -0,0 +1,34 @@
...
+  moduleServer(
+    id,
+    function(input, output, session) {
+      output$invalid_model <- renderUI({
+        if (model_valid()) {
+          return()
+        } else {
+          return(
+            div(
+              h3("Please rerun model"),
```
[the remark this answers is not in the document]
JanionNevill: It was a little awkward, but it works now

JanionNevill:
Also updated a load of Roxygen comments

R/bayesian/model_details_panel.R
```diff
@@ -108,17 +115,45 @@ model_details_panel_ui <- function(id, item_names, page_numbering) {
...
     
     main_model <- models[[1]]
+    main_model_valid <- models_valid[[1]]
+    
+    invalid_model_panel_server(id = "model_invalid_1", model_valid = main_model_valid)
+    invalid_model_panel_server(id = "model_invalid_2", model_valid = main_model_valid)
+    invalid_model_panel_server(id = "model_invalid_3", model_valid = main_model_valid)
+    
+    observe({
+      # Get either enble or disable function to apply to relevant UI elements
```
nabury, line 133: Typo

nabury (changes requested):
Looks good - could you change "please rerun model" to "please run model" when a new dataset is used as discussed on email?

nabury (changes requested):
Just noticed all the instances of "Please note: if you change the selections on the sidebar, you will need to re-run the primary and/or sensitivity analysis from the 'Forest Plot' page." can now be removed.

nabury (approved):
👍

PR #217, opened 2025-01-20
Deploy action workflow

JanionNevill:
Copied the deploy action from metaImpact, with slight changes to point to this app, and to add an optional field to select the branch to deploy.


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,97 @@
...
+  workflow_dispatch:
+    inputs:
+      deploy_type:
+        description: Deploy to release or test
+        required: true
+        options:
+          - 'test'
+          - 'release'
+        default: 'test'
+      manual_branch:
```
RyanJField, line 18: This should not be necessary, by default once the workflow has been pushed to main (or run once), workflow dispatch will be enabled and ask you which branch you want to use, see MetaBayesDTA:
https://github.com/CRSU-Apps/MetaBayesDTA/blob/main/.github/workflows/deploy.yaml#L9
https://github.com/CRSU-Apps/MetaBayesDTA/actions/workflows/deploy.yaml

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,97 @@
...
+          rsconnect::deployApp(
+            appName = Sys.getenv("appName"),
+            account = Sys.getenv("accountName"),
+            forceUpdate = TRUE
+          )
+        shell: Rscript {0}
+        env:
+          accountName: crsu
+          # token and secret obtained from https://www.shinyapps.io/admin/#/tokens
+          accountToken: ${{ secrets.SHINYAPPS_TOKEN }}
```
RyanJField, line 96: These need to be set in repository secrets before use, settings > secrets and variables > repository secrets

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,97 @@
...
+          # Store the appName in environment variables based on deployment type
+          if [[ $DEPLOY_TYPE = "test" ]]; then
+            echo "appName=MetaInsight_test" >> $GITHUB_ENV
+          else
+            echo "appName=MetaInsight" >> $GITHUB_ENV
+          fi
+      - name: Print deployment type
+        run: |
+          echo "Deploying to: $appName"
+      - name: Install gcc 11
```
RyanJField, line 50: This may not be necessary for MetaInsight as it doesn't require STAN.

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - uses: actions/checkout@v4
+        with:
+          ref: "{{github.event.inputs.manual_branch}"
+      # Setup R (4.3.0)
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          r-version: '4.3.2'
+      # Setup and install dependencies using renv
+      - uses: r-lib/actions/setup-renv@v2
+      - name: Check GCC Version
```
RyanJField, line 68: See previous comment

RyanJField (changes requested):
Remove manual branch (see comments), or move checkout earlier so that it fails sooner if the branch does not exist.

JanionNevill:
I realise that I branched off `dev` instead of `main`, so I'm going to close this PR, and reopen on a new branch off `main`.

PR #218, opened 2025-01-20
Create deploy.yml

JanionNevill:
I've removed the GCC and branch name aspects. I'll add the secrets momentarily.


RyanJField (approved):
Looks good.

RyanJField:
One note, you may need to call renv::hydrate prior to deploying as MetaInsight does not have and Renv lock file.

PR #219, opened 2025-01-27
Created lockfile before deployment

JanionNevill:
Yes, this was needed. I ran the workflow and it failed due to a missing lockfile. The documentation suggested that `renv::snapshot()` was more appropriate than `renv:;hydrate()`


RyanJField:
Have you tested this, now that the workflow has been run once, you can run it on this branch.

RyanJField:
Ren may not be necessary, as I think deploying will capture the dependencies, but I'm not 100%

PR #221, opened 2025-01-29
Version 6.2.1

tommorris168:
Also added the names of the bnma developers, which I forgot in the last task.


JanionNevill (approved):
Looks good

PR #222, opened 2025-02-04
Deploy workflow 2

JanionNevill:
To get the workflow to function properly, I had to `{renv}` the project. We had always intended on doing that anyway though.


.github/workflows/deploy.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           echo "Deploying to: $appName"
       - name: Install package system dependencies
         run: |
-          sudo apt install libcurl4-openssl-dev jags libharfbuzz-dev libfribidi-dev libfreetype6-dev libpng-dev libtiff5-dev libjpeg-dev libglpk-dev -y
+          sudo apt install libcurl4-openssl-dev jags libharfbuzz-dev libfribidi-dev libfreetype6-dev libpng-dev libtiff5-dev libjpeg-dev libglpk-dev libmagick++-dev -y
       # Checkout the repo
       - uses: actions/checkout@v4
       # Setup R (4.3.0)
```
RyanJField, line 50: Ideally, this would match the R version being used (4.3.2)

RyanJField (approved):

.github/workflows/deploy.yml
JanionNevill: Updated here and in the lockfile

PR #223, opened 2025-02-04
Mi 50 remove hard coding of 6 arms

tommorris168:
I have uploaded a data set with seven arms to the test folder. It isn't used in the tests, it's for you to run.


tommorris168:
It turns out that studlab is now mandatory even for wide data, despite the help saying it is mandatory only for long data.

JanionNevill:
I get the following warnings after updating the `{netmeta}` and `{meta}` packages:
```
Warning: Use argument 'level.ma' instead of 'level.comb' (deprecated).
Warning: Use argument 'common' instead of 'comb.fixed' (deprecated).
Warning: Use argument 'random' instead of 'comb.random' (deprecated).
```

The offending code is in `freq.df()` in `fn_analysis.R`. While you're there, could you `::` scope the `netmeta()` function please?

JanionNevill:
Can you add `library(meta)` to `global.R` please?

JanionNevill (changes requested):
Everything seems to be working. Just a couple of minor changes.
Good job 👍

tommorris168:
I've also snook in two more changes:
There was a note from months ago to remove an unused parameter in freq.df(). It is now removed.
The forest plot on tab 1b was cutting off the bottom study, so I've changed the y-axis minimum from 1 to 0.

JanionNevill (approved):
✔

PR #224, opened 2025-02-10
Dev workflow 2

JanionNevill:
Added `sudo apt-get update` before installing package dependencies, otherwise the build was failing

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,7 +44,7 @@ jobs:
           echo "Deploying to: $appName"
       - name: Install package system dependencies
         run: |
-          sudo apt-get update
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -44,6 +44,7 @@ jobs:
           echo "Deploying to: $appName"
       - name: Install package system dependencies
         run: |
+          sudo apt-get update
```
RyanJField, line 47: Please stick to either `apt` or `apt-get`, the next line is using `apt`. `apt` should be the default package manager for linux distros.

RyanJField (changes requested):

RyanJField (approved):

PR #227, opened 2025-03-10
Version 6.3.0

tommorris168:


tommorris168:
@JanionNevill Is there anything you want to add to the list of changes about the hiding?

tommorris168:
@JanionNevill I just had to make another commit because some of the parameters in CreateCompositeMetaRegressionPlot() had been reverted. I don't know how this could have happened, as I've still got the original branch, and it was right there. Very strange...

R/home_page.R
```diff
@@ -35,21 +35,16 @@ home_page_ui <- function(id) {
...
-        p(
-          tags$ul(
-            tags$li(
-              "Meta-regression has been added. One covariate is allowed, which can be a new continuous or binary variable,
-              or baseline risk. Two new graphs are available for meta-regression. The first displays the covariate values grouped by
-              treatment and study. The second plots the covariate against relative treatment effects, with credible regions and
-              study-level contributions."
-            )
-          )
-        ),
```
JanionNevill, line 52: I think that this should list the updates from v6.0.0 and v6.3.0, not v6.2.1

We keep the most recent major and minor releases, not just the 2 most recent releases

R/home_page.R
```diff
@@ -35,21 +35,16 @@ home_page_ui <- function(id) {
       column(
         width = 5,
         p(tags$strong("Latest Updates:")),
+        p(tags$strong("Minor update (11 March 2025 v6.3.0):")),
+        tags$ul(
+          tags$li("Removed the limit of 6 arms per study in the frequentist analysis."),
+          tags$li("Removed the sizing of circles by contribution in the regression plot."),
+          tags$li("Fixed a bug in the study-level forest plot where the last treatment effect was partially cut off."),
+        ),
```
JanionNevill, line 43: Please add:
"Bayesian analysis outputs are now cleared when options are changed"

JanionNevill (changes requested):
One update description to add, one change to previous updates.

R/home_page.R
tommorris168: Thanks, I wasn't sure about that.

JanionNevill (approved):
✔

PR #229, opened 2025-03-26
Cin-mi-11-add-qa-variables

tommorris168:
I think we should get the approval of others before closing this PR, as they might have an opinion on the placement of the circles.

### DIFF
--- a/R/data/data_definitions.R
+++ b/R/data/data_definitions.R
@@ -12,7 +12,7 @@
   type_check = "is.character",
   type_name = "character (text)",
   pattern = "^(?i)Study(\\.([0-9]+))?$",
-  replacement ="Study\\1",
+  replacement = "Study\\1",
   number_group = NA
 )
 
@@ -22,7 +22,7 @@
   type_check = "is.character",
   type_name = "character (text)",
   pattern = "^(?i)T(\\.([0-9]+))?$",
-  replacement ="T\\1",
+  replacement = "T\\1",
   number_group = "\\2"
 )
 
@@ -32,7 +32,7 @@
   type_check = "is.numeric",
   type_name = "numeric",
   pattern = "^(?i)N(\\.([0-9]+))?$",
-  replacement ="N\\1",
+  replacement = "N\\1",
   number_group = "\\2"
 )
 
@@ -42,7 +42,7 @@
   type_check = "is.numeric",
   type_name = "numeric",
   pattern = "^(?i)Mean(\\.([0-9]+))?$",
-  replacement ="Mean\\1",
+  replacement = "Mean\\1",
   number_group = "\\2"
 )
 
@@ -52,7 +52,7 @@
   type_check = "is.numeric",
   type_name = "numeric",
   pattern = "^(?i)SD(\\.([0-9]+))?$",
-  replacement ="SD\\1",
+  replacement = "SD\\1",
   number_group = "\\2"
 )
 
@@ -62,7 +62,7 @@
   type_check = "is.numeric",
   type_name = "numeric",
   pattern = "^(?i)covar\\.(.+)$",
-  replacement ="covar.\\1",
+  replacement = "covar.\\1",
   number_group = NA
 )
 
@@ -72,8 +72,28 @@
   type_check = "is.numeric",
   type_name = "numeric",
   pattern = "^(?i)R(\\.([0-9]+))?$",
-  replacement ="R\\1",
+  replacement = "R\\1",
   number_group = "\\2"
+)
+
+.rob_definition <- data.frame(
+  name = "rob",
+  required = FALSE,
+  type_check = "is.numeric",
+  type_name = "numeric",
+  pattern = "^(?i)rob$",
+  replacement = "rob\\1",
+  number_group = NA
+)
+
+.indirectness_definition <- data.frame(
+  name = "indirectness",
+  required = FALSE,
+  type_check = "is.numeric",
+  type_name = "numeric",
+  pattern = "^(?i)indirectness$",
+  replacement = "indirectness\\1",
+  number_group = NA
 )
 
 continuous_column_names <- data.frame() %>%
@@ -82,11 +102,15 @@ continuous_column_names <- data.frame() %>%
   rbind(.n_definition) %>%
   rbind(.mean_definition) %>%
   rbind(.sd_definition) %>%
-  rbind(.covariate_definition)
+  rbind(.covariate_definition) %>%
+  rbind(.rob_definition) %>%
+  rbind(.indirectness_definition)
 
 binary_column_names <- data.frame() %>%
   rbind(.study_definition) %>%
   rbind(.t_definition) %>%
   rbind(.r_definition) %>%
   rbind(.n_definition) %>%
-  rbind(.covariate_definition)
+  rbind(.covariate_definition) %>%
+  rbind(.rob_definition) %>%
+  rbind(.indirectness_definition)
--- a/R/data/data_validation.R
+++ b/R/data/data_validation.R
@@ -298,7 +298,7 @@ ValidateUploadedData <- function(data, outcome_type) {
 
   studies_with_wrong_qa_values <- unique(data$Study[!(data$rob %in% 1:3) | !(data$indirectness %in% 1:3)])
   if (!is.null(data$rob) && !is.null(data$indirectness)
-      && length(studies_with_wrong_qa_values) != 0) {
+      && length(studies_with_wrong_qa_values) > 0) {
     return(
       list(
         valid = FALSE,
--- a/R/data/data_validation.R
+++ b/R/data/data_validation.R
@@ -50,6 +50,11 @@ ValidateUploadedData <- function(data, outcome_type) {
     return(result)
   }
 
+  result <- .ValidateQualityColumns(data)
+  if (!result$valid) {
+    return(result)
+  }
+  
   return(.valid_result)
 }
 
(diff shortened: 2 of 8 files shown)


R/data/data_validation.R
```diff
@@ -266,5 +271,61 @@ ValidateUploadedData <- function(data, outcome_type) {
...
+  if (length((studies_with_multiple_rob) > 0)) {
+    return(
+      list(
+        valid = FALSE,
+        message = glue::glue("Some studies do not have a unique risk of bias: {paste0(studies_with_multiple_rob, collapse = ', ')}.")
+      )
+    )
+  }
+  
+  studies_with_multiple_indirectness <- unique(data$Study)[as.vector(rowSums(table(data[, c("Study", "indirectness")]) != 0) > 1)]
```
JanionNevill, line 320: This might need a comment to explain a fairly dense line of code

R/data/data_validation.R
```diff
@@ -266,5 +271,61 @@ ValidateUploadedData <- function(data, outcome_type) {
...
+  if (length(studies_with_multiple_indirectness) > 0) {
+    return(
+      list(
+        valid = FALSE,
+        message = glue::glue("Some studies do not have a unique indirectness: {paste0(studies_with_multiple_indirectness, collapse = ', ')}.")
```
JanionNevill, line 325: This message could be misinterpreted. Suggested message:
"Some studies do not have the same indirectness value for every arm:..."

R/study_exclusions_panel.R
```diff
@@ -29,18 +29,63 @@ study_exclusions_panel_ui <- function(id) {
...
+      }
+    })
+    
     data_reset <- reactiveVal(FALSE)
     
+    #' Get the colour corresponding to the quality assessment values.
+    #' 
+    #' @param quality_value 1, 2 or 3.
+    #' @return Named vector of 'icon' and 'colour'.
+    QualityTagColour <- function(quality_value) {
```
JanionNevill, line 51: Can you factor this out as a private function (with "." prefixing the function name) in this file, but outside the module please?

tests/testthat/test-data_validation.R
```diff
@@ -239,4 +239,42 @@ test_that("ValidateUploadedData() identifies invalid binary long data, with sing
...
+  validation_result <- ValidateUploadedData(data, "Continuous")
+  validation_result2 <- ValidateUploadedData(data2, "Continuous")
+  
+  expect_false(validation_result$valid)
+  expect_equal(validation_result$message, "Provide both 'rob' and 'indirectness', or neither.")
+  expect_false(validation_result2$valid)
+  expect_equal(validation_result2$message, "Provide both 'rob' and 'indirectness', or neither.")
+})
+
+test_that("ValidateUploadedData() identifies unallowed quality assessment values", {
```
JanionNevill, line 260: Can you also test the same check for indirectness please? Same in the next test case.

tests/testthat/test-data_validation.R
JanionNevill, line 1: Can you add test cases for checking when the rob and indirectness columns are partially populated please? ie. with some NAs thrown in

JanionNevill, line 1: Can you add a test case for a valid data set with RoB and indirectness included please?

tests/testthat/test-data_wrangling.R
JanionNevill, line 1: Can you add some test cases for the case insensitivity of the new columns please? See line 206 onwards

JanionNevill (changes requested):
I haven't been able to run the tests because the renv versionm of netmeta is causing problems again. Did you change the version when you overwrote the lock file recently?

R/data/data_validation.R
```diff
@@ -266,5 +271,61 @@ ValidateUploadedData <- function(data, outcome_type) {
...
+#' @return Validation result in the form of a list:
+#' - "valid" = TRUE or FALSE defining whether data is valid
+#' - "message" = String describing any issues causing the data to be invalid
+.ValidateQualityColumns <- function(data) {
+  if (is.null(data$rob) && is.null(data$indirectness)) {
+    return(.valid_result)
+  }
+  
+  if ((is.null(data$rob) && !is.null(data$indirectness))
+      || (!is.null(data$rob) && is.null(data$indirectness))) {
```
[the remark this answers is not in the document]
tommorris168: Excellent, I thought that code was quite ugly. I'll go with xor.

R/data/data_validation.R
```diff
@@ -266,5 +271,61 @@ ValidateUploadedData <- function(data, outcome_type) {
...
+    return(
+      list(
+        valid = FALSE,
+        message = glue::glue("Provide both 'rob' and 'indirectness', or neither.")
+      )
+    )
+  }
+
+  studies_with_wrong_qa_values <- unique(data$Study[!(data$rob %in% 1:3) | !(data$indirectness %in% 1:3)])
```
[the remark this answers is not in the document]
tommorris168: I think I have to use `|` here because the two arguments are vectors, and `||` only works on scalars.

"Using vectors of more than one element in `&&` or `||` will give an error." from https://stat.ethz.ch/R-manual/R-patched/library/base/html/Logic.html

tommorris168:
I had not updated renv. I've taken the two files that were updated (renv.lock and activate.R) from my local copy of main and put them in this branch. I hope that works.

R/data/data_validation.R
JanionNevill: Is there a test case which could be added which would match this case?

JanionNevill (changes requested):
A couple of minor comments left.

Renv is now working. Your changes didn't instantly fix the issue, so I deactivated the project then reactivated it. That worked, but I don't know whether it would have worked without your changes...

R/data/data_validation.R
tommorris168: It's already there, on lines 321 and 322 of test-data_validation.

R/study_exclusions_panel.R
```diff
@@ -1,4 +1,17 @@
 
+#' Get the colour corresponding to the quality assessment values.
+#' 
+#' @param quality_value 1, 2 or 3.
+#' @return Named vector of 'icon' and 'colour'.
```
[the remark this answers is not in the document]
tommorris168: Changed to say the colour is CSS text.

The icon bit was left over from when this function also returned an icon, which I got rid of in favour of the icon always being a circle. I attempted to have icons with a plus, minus or cross as in the robvis package, but I couldn't get the cross to be red. I could only get its border to be red.
https://cran.r-project.org/web/packages/robvis/vignettes/Introduction_to_robvis.html

R/data/data_validation.R
JanionNevill: Only one of those checks fails with the old incorrect code. Changing line 311 to this would be a more comprehensive test case:
```data$rob <- c(1, 1, 1, 2, 2, 2, 1, 3)```

File: R/study_exclusions_panel.R (hunk 7 above)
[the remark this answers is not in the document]
Comment (JanionNevill) [reply R/study_exclusions_panel.R]: Ah, I'm with you now. Did you try using other CSS properties other than "color"? It may have needed something like `bavkground-color`.

Would you mind adding an explicit `return()` call around the statement please?

**Review by JanionNevill — COMMENTED** [review]
Minor comments. Sorry!

File: R/data/data_validation.R (hunk 1 above)
Comment (tommorris168) [reply R/data/data_validation.R]: I'm not sure what you mean. If both `rob` and `indirectness` have multiple values in the same study, then an error message will only be given for `rob`, because it is checked first and if an error is found the checks stop. The two cases currently have an error in `rob` and an error in `indirectness`. Are you asking for a third test case where both have an error?
Your suggested change on line 311 is only moving the error from Constantine to Justinian. The first three values are for Constantine and are currently 1, 1, 2.

File: R/study_exclusions_panel.R (hunk 7 above)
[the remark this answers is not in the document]
Comment (tommorris168) [reply R/study_exclusions_panel.R]: I did try some other things but I gave up after a while. I may try again before we merge the PR.

**Comment by tommorris168** [comment]
We'll leave this unmerged until Clareece, Naomi, Alex and Nicola have all had their say. Clareece and Naomi can see it on Shiny Friday, Alex and Nicola at my catch-up on Monday.

**Comment by tommorris168** [comment]
@JanionNevill I've done the right-alignment, although it might not be an optimal solution. I fiddled around with the values in the  length function until alighting on 7 pixels per character in the study name and 100 pixels for the quality variables. It works for various study name lengths.

I've also now got the icons working. It needed the `class = "fa-solid"` argument to make the red X work. The strange thing is that this argument was already there, so I must have added it for the circles and then not tested it with the other icons.

I'll wait until I've spoken to Alex and Nicola on Monday before requesting another review.

File: R/data/data_validation.R [hunk 8]
```diff
@@ -282,52 +282,59 @@ ValidateUploadedData <- function(data, outcome_type) {
 #' - "valid" = TRUE or FALSE defining whether data is valid
 #' - "message" = String describing any issues causing the data to be invalid
 .ValidateQualityColumns <- function(data) {
-  if (is.null(data$rob) && is.null(data$indirectness)) {
+  #The rob and indirectness columns
+  rob_indirectness_columns <- grep(pattern = "^rob|^indirectness$", x = names(data), value = TRUE)
```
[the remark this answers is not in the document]
tommorris168: I'm not convinced. I do want `rob_indirectness_columns` for the loop below, so splitting it up seems like it would create more code.

R/data/data_definitions.R
```diff
@@ -81,7 +81,7 @@
   required = FALSE,
   type_check = "is.numeric",
   type_name = "numeric",
-  pattern = "^(?i)rob$",
+  pattern = "^(?i)rob(\\.)?.*$",
   replacement = "rob",
```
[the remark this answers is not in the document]
JanionNevill: I just had a feeling, so I added a `browser()` call at the end of `WrangleData()` to check what was coming out.

R/data/data_validation.R
```diff
@@ -282,52 +282,59 @@ ValidateUploadedData <- function(data, outcome_type) {
...
+  #The number of individual RoB columns
+  n_rob_individual <- length(grep(pattern = "^rob.+$", x = names(data)))
+
+  #Check there are no more than 10 individual RoB columns
+  if (n_rob_individual > 10) {
     return(
       list(
         valid = FALSE,
-        message = glue::glue("Provide both 'rob' and 'indirectness', or neither.")
+        message = glue::glue("A maximum of 10 individual risk of bias variables are allowed.")
```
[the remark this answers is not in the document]
JanionNevill: I think that this is a safe number. We can adjust it later if needed, but I doubt it will be.

R/data/data_validation.R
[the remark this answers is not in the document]
JanionNevill: It would be slightly more code, but I think that it's best not to evaluate complex code within an if statement where possible, like on line 306. Not important, just a suggestion.

JanionNevill (commented):
✔

PR #230, opened 2025-04-14
Mi 95 add trace and density plots

tommorris168:
I couldn't get `MCMCvis::MCMCtrace()` to work, so I used `bayesplot::mcmc_trace()` and `bayesplot::mcmc_hist()` instead. They return a ggplot so the code to arrange them is different to the Gelman plots.

There are a couple of things I'm not happy with:

- The plots don't use the full width of the screen in the regression models. The same is true for NMA but that's fine because the sensitivity model takes up half the screen.
- They take a long time to run. These plots are actually simple to make, so at some point in the near future I'd like to draw them manually and see if they run any faster.

Since we're on a deadline to resubmit the paper (28th April, seven working days) I do not want to address either of the above in this task.

### DIFF
--- a/R/bayesian/deviance/mcmc_panel.R
+++ b/R/bayesian/deviance/mcmc_panel.R
@@ -9,19 +9,21 @@ mcmc_panel_ui <- function(id, item_name) {
   div(
     invalid_model_panel_ui(id = ns("model_invalid")),
 
-    p(tags$strong(glue::glue("Gelman convergence assessment plots for {item_name}"))),
+    h3(tags$strong(glue::glue("Gelman convergence assessment plots for {item_name}"))),
     shinycssloaders::withSpinner(
       plotOutput(outputId = ns("gemtc_gelman"), inline = TRUE),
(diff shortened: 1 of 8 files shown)


R/bayesian/deviance/mcmc_panel.R
```diff
@@ -0,0 +1,159 @@
...
+    )
+    
+    #Trace plots
+    output$trace_plots <- renderPlot(
+      {
+        if (is.null(model_valid()) || !model_valid()) {
+          return()
+        }
+        
+        plotlist <- reactive(
```
JanionNevill, line 116: This shouldn't be a reactive, it's already in a reactive context by being in the `renderPlot()` call.

Same on line 139

R/plot.R
JanionNevill, line 1: The MCMC plots are very narrow for meta-regression. Can these be resized depending on whether there is a sensitivity analysis or not?

JanionNevill (approved):
✔

PR #231, opened 2025-04-15
Mi 96 add bayesian details

tommorris168:
I've added download buttons, which wasn't on Jira.
I have not done a test for the prior distributions in gemtc, because the values are reactive to the data and I don't know how they are chosen.
The same might actually be true of bnma, so perhaps that test isn't worth much.

It was tricky to work out the prior distributions. You have to look at the model code and then extract the relevant values from the data. For example, in gemtc the relative treatment effects have distribution `dnorm(0, prior.prec)`. prior.prec is defined by `prior.prec <- pow(re.prior.sd, -2)`, and `re.prior.sd` is not defined in the code, which means it's a value passed to JAGS. Once `re.prior.sd` was extracted from the model, I put the Normal distribution in its usual form, which is N(mean, variance). For some reason in Bayesian programs they use N(mean, precision) instead, where precision = 1/variance. In this case the variance is `re.prior.sd^2`.
When we first started meta-regression I did a comparison of prior distributions between gemtc and bnma, which was immensely helpful for this task. It's attached in case it helps you.

Since this is quite statsy, you could just review the rest of the code and we could find someone else to check the priors, if you're not happy doing it.

### DIFF
--- a/R/bayesian/model_details_panel.R
+++ b/R/bayesian/model_details_panel.R
@@ -34,15 +34,30 @@ model_details_panel_ui <- function(id, item_names, page_numbering) {
   ui = div(
     tabsetPanel(
       tabPanel(
-        title = paste0(page_numbering$AddChild(), " Model codes"),
+        title = paste0(page_numbering$AddChild(), " MCMC characteristics and priors"),
         invalid_model_panel_ui(id = ns("model_invalid_1")),
+        downloadButton(outputId = ns('download_mcmc'),
+                       label = "Download MCMC"),
+        downloadButton(outputId = ns('download_priors'),
+                       label = "Download priors"),
+        h4(tags$strong("MCMC characteristics")),
+        tableOutput(outputId = ns("mcmc_details")),
+        h4(tags$strong("Prior distributions")),
+        tableOutput(outputId = ns("priors")),
+        br(),
+        p("Note: Normal distributions are parameterized here as N(mean, variance) and in the JAGS code as N(mean, precision)."),
+        p("Note: For help on the scaled t-distribution see the JAGS manual.")
+      ),
+      tabPanel(
+        title = paste0(page_numbering$AddChild(), " Model codes"),
+        invalid_model_panel_ui(id = ns("model_invalid_2")),
         p(tags$strong("Model codes for analysis of all studies")),
         downloadButton(outputId = ns('download_code')),
         verbatimTextOutput(outputId = ns("code"))
       ),
       tabPanel(
         title = paste0(page_numbering$AddChild(), " Initial values"),
-        invalid_model_panel_ui(id = ns("model_invalid_2")),
+        invalid_model_panel_ui(id = ns("model_invalid_3")),
         p(tags$strong("Initial values")),
         downloadButton(outputId = ns('download_inits_1'), "Download initial values for chain 1"),
         downloadButton(outputId = ns('download_inits_2'), "Download initial values for chain 2"),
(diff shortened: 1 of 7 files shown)


R/data/data_wrangling.R
```diff
@@ -543,4 +543,19 @@ FindMaxArms <- function(data) {
...
   }
+}
+
+
+
+#' Returns a number rounded in a nice manner.
+#' 
+#' @param variance Any number, doesn't have to be a variance.
+#' @return 'variance' with 1 decimal place or 1 significant figure.
+RoundVariance <- function(variance) {
```
JanionNevill, line 554: If this doesn't have to be variance, can it get a more representative name please? Maybe "RoundForDisplay"?

R/meta_regression/bnma_analysis.R
```diff
@@ -506,3 +505,81 @@ BnmaRelativeEffects <- function(model, covariate_value) {
...
+                                              paste0(" ~ Unif (", exchangeable_sd_lower, ", ", exchangeable_sd_upper, ")")))
+    )
+    #If the model is NMR unrelated, add the covariate parameters
+  } else if (model$network$baseline == "independent") {
+    unrelated_mean <- model$network$prior.data$mean.bl
+    unrelated_var <- round(1 / model$network$prior.data$prec.bl, digits = 1)
+    prior_table <- rbind(prior_table,
+                         data.frame(parameter = "Covariate parameters",
+                                    value = paste0(" ~ N (", unrelated_mean, ", ", unrelated_var, ")"))
+    )
```
JanionNevill, line 582: A lot of function calls are not conforming to the [style guide that we have adopted](https://style.tidyverse.org/syntax.html#long-function-calls).

I know it's annoying to redo them all to perfectly match the guide, but could the closing brackets for function calls be moved such that it's clear where the arguments end please?
 eg.
```
    prior_table <- rbind(prior_table,
                         data.frame(parameter = "Covariate parameters",
                                    value = paste0(" ~ N (", unrelated_mean, ", ", unrelated_var, ")")
                                   )
                        )
```
Ideal would be
```
    prior_table <- rbind(
        prior_table,
        data.frame(
            parameter = "Covariate parameters",
            value = paste0(" ~ N (", unrelated_mean, ", ", unrelated_var, ")")
        )
    )
```

This applies to multiple files on this task.

JanionNevill (changes requested):
Just some formatting and hooking up

R/bayesian/model_details_panel.R
```diff
@@ -34,15 +34,30 @@ model_details_panel_ui <- function(id, item_names, page_numbering) {
   ui = div(
     tabsetPanel(
       tabPanel(
-        title = paste0(page_numbering$AddChild(), " Model codes"),
+        title = paste0(page_numbering$AddChild(), " MCMC characteristics and priors"),
         invalid_model_panel_ui(id = ns("model_invalid_1")),
+        downloadButton(outputId = ns('download_mcmc'),
+                       label = "Download MCMC"),
+        downloadButton(outputId = ns('download_priors'),
+                       label = "Download priors"),
```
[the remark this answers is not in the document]
tommorris168: For consistency with the model details tab. All subtabs have the download buttons at the top. (Why that was done in the first place, don't ask me).

R/meta_regression/bnma_analysis.R
tommorris168: I've made various changes to the whole file. I think it matches the guide now.

R/bayesian/model_details_panel.R
```diff
@@ -177,20 +180,22 @@ model_details_panel_server <- function(id, models, models_valid, package = "gemt
...
     
-    output$mcmc_details <- renderTable({
-      mcmc_details()
-    },
-    digits = 0,
-    colnames = FALSE
+    output$mcmc_details <- renderTable(
+      {
+        mcmc_details()
+      },
```
JanionNevill, line 186: This table and the priors table are not hiding when the model becomes invalid. This should follow the pattern on line 216

JanionNevill (changes requested):
The tables are not hiding as they should, and one minor tidy-up

JanionNevill (approved):
All good. Thanks, Tom

PR #239, opened 2025-07-03
CZI json schema

JanionNevill:
This is the backend code for exporting the frequentist analysis into a json format that CINeMA can read. I have also added data definitions, but these will be superseded by the work done on https://github.com/CRSU-Apps/MetaInsight/pull/229.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -51,7 +51,7 @@ jobs:
       # Setup R (4.3.2)
       - uses: r-lib/actions/setup-r@v2
         with:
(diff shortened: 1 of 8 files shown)


tommorris168 (changes requested):
A few minor comments.
The json code looks right though I don't know the syntax. The fact that it's accepted by jsonvalidate::json_validate() is reassuring.

JanionNevill:
I've addressed your comments here, and I've continued development which will be best placed into a separate PR

tommorris168 (approved):
Looks good

PR #242, opened 2025-09-12
cin-mi-14-allow-rob-and-indirectness-in-upload

tommorris168:
The changes are almost identical to those in Cin-mi-11-add-qa-variables:
https://github.com/CRSU-Apps/MetaInsight/pull/229

There are two differences:

In the UI where it is explained what variables the uploaded dataset could/should contain, I have not included a link to CINeMA as the code that worked previously didn't work in the new markdown file.

The bits that involve `.ContinuousOrder()` and `.BinaryOrder()` have not been included, because these functions are not there. They were introduced in Mi-50-remove-hard-coding-of-6-arms:
https://github.com/CRSU-Apps/MetaInsight/pull/223
The code is not currently in the Scholar version, so will have to be added at a later date.

### DIFF
--- a/inst/shiny/modules/setup_load.md
+++ b/inst/shiny/modules/setup_load.md
@@ -90,13 +90,13 @@ numeric treatment IDs in the "T" column(s) with the treatment names.
 <ul>
 <li>
 <strong>rob</strong>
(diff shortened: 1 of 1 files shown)


simon-smart88:
This all looks good to me. I've fixed the links to cinema and made a note to ensure that those changes in #223 are carried over at some point.

PR #243, opened 2025-11-05
CINeMA Schema Improvements

JanionNevill:
Made changes to the CINeMA export inline with Thodoris' suggestions.

There are 2 skipped tests:
- One validating missing fields in the schema. This seems to be due to [a bug in the {jsonvalidate} package](https://github.com/ropensci/jsonvalidate/issues/85).
- One checking the recreation of the study contributions. This is likely due to _CINeMA_ using an older version of {netmeta}

### DIFF
--- a/tests/testthat/test-cinema_export.R
+++ b/tests/testthat/test-cinema_export.R
@@ -69,8 +69,6 @@ RecreateAnalysisFromCinemaProject <- function() {
     outcome_type = cinema_outcome_type
   )
   
-  cinema_project <<- cinema_project
-  
   calculated_analysis <- GenerateCinemaAnalysis(data, cinema_model_type, cinema_outcome_measure)
   
   exported_json <- GenerateCinemaJson(


tests/testthat/test-cinema_export.R
```diff
@@ -47,16 +48,157 @@ GenerateCinemaAnalysis <- function(cinema_data, model_type, outcome_measure) {
...
+    CleanTreatmentIds()
+  cinema_long_data <- WrangleUploadData(cinema_long_data, treatment_ids, cinema_outcome_type)
+  
+  data <- list(
+    long_data = cinema_long_data,
+    treatment_ids = treatment_ids,
+    outcome_type = cinema_outcome_type
+  )
+  
+  cinema_project <<- cinema_project
```
tommorris168: Is this line necessary? Since you return `cinema_project` anyway it looks like it isn't doing anything.

PR #246, opened 2026-01-12
add v7 banner

simon-smart88:
As discussed, I've added a banner at the top of the home page pointing to the beta version of v7. I will add an equivalent to v7 once we switch them around. Example below (link opens in new window) 

image

### DIFF
--- a/R/home_page.R
+++ b/R/home_page.R
@@ -24,7 +24,9 @@ home_page_ui <- function(id) {
     fluidRow(
       div(
         h4("A Beta version of MetaInsight v7 featuring improved plots, automatically updating models and other new features is available at: ",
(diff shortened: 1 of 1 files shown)


R/home_page.R
```diff
@@ -21,6 +21,16 @@ home_page_ui <- function(id) {
         });'
       )
     ),
+    fluidRow(
+      div(
+        h4("A Beta version of MetaInsight v7 featuring improved plots, automatically updating models and other new features is available at: ",
+           a("https://crsu.shinyapps.io/MetaInsight_Scholar/", href= "https://crsu.shinyapps.io/MetaInsight_Scholar/", target = "_blank", style="color: white"),
```
JanionNevill: Can you add an underline to this as well please? It isn't totally obvious that it's a link at the moment.

JanionNevill (approved):
Looks good

JanionNevill (approved):
✔

PR #254, opened 2026-02-18
Czi json export scholar

tommorris168:
I believe I've done everything except end-to-end tests, where I wasn't sure what to do.

Much of this was written by Janion and has already been reviewed by me, although I did have to edit some of it in light of the recent changes to remove duplicate parameters.


inst/shiny/modules/rep_cinema.R
```diff
@@ -0,0 +1,108 @@
...
+      shinyjs::hide("download")
+    }
+  })
+
+  gemtc_results <- reactive({
+    if (input$select_model == "freq") {
+      return(NULL)
+    } else if (input$select_model == "bayes") {
+      if (input$select_data == "full") {
+        return(common$bayes_all$mtcResults)
```
simon-smart88: this can just return `"bayes_all"` or `"bayes_sub"` and then use `gemtc = common[[gemtc_results()]]` in the function call

inst/shiny/modules/rep_cinema.R
```diff
@@ -0,0 +1,108 @@
+rep_cinema_module_ui <- function(id) {
+  ns <- shiny::NS(id)
+  tagList(
+    # UI
+
+    radioButtons(ns("select_model"), "Select model to export", choices = list("Frequentist" = "freq", "Bayesian" = "bayes")),
+    radioButtons(ns("select_data"), "Select dataset", choices = list("Full" = "full", "Sensitivity" = "sensitivity")),
```
simon-smart88: I think this would be more consistent and then the input value can be passed directly to the function: `choices = list("All studies" = "configured_data", "With selected studies excluded" = "subsetted_data"))`

tests/testthat/test-rep_cinema.R
```diff
@@ -0,0 +1,220 @@
...
+
+
+
+
+
+
+
+
+test_that("{shinytest2} recording: e2e_rep_cinema", {
+  app <- shinytest2::AppDriver$new(app_dir = system.file("shiny", package = "metainsight"), name = "e2e_rep_cinema")
```
simon-smart88: let's wait to add this until after we merge as I've just refactored all the tests and we'll just get confused

R/rep_cinema_f.R
```diff
@@ -0,0 +1,479 @@
...
+#' This is a named list of lists. The top level list contains items:
+#' - "project" Information for CINeMA project
+#'   - "CM" Contribution matrices
+#'     - "contributionMatrices" output from `.PrepareAnalysisForCinema()`
+#'   - "format" Data format. Always "long"
+#'   - "type" Outcome type. Either "binary" or "continuous"
+#'   - "Studies" Study data
+#'     - "long" Output from `.PrepareDataForCinema()`
+#' 
+#' @param data The dataframe used in MetaInsight.
```
simon-smart88: Once refactored you can use `@inheritParams common_params` to document `configured_data` but the `model` paramter will need documenting.

simon-smart88 (commented):
Hi Tom

I've made various comments - mostly just about reorganising things to be more consistent and a bit simpler. I think the tests might need some refactoring for consistency with the rest, but I'm happy to take care of that after merging.

Let me know if anything's unclear and happy to chat through if it would be helpful.

Simon

inst/shiny/modules/rep_cinema.R
```diff
@@ -0,0 +1,108 @@
...
+    filename = function() {
+      paste0("cinema_", input$select_model, ".json")
+    },
+    content = function(file) {
+      writeLines(cinema(), file)
+    })
+
+  return(list(
+    save = function() {
+      # Save any values that should be saved when the current session is saved
```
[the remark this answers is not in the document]
tommorris168: I don't know what this bit is for. Is it for populating common? Is there anything that needs saving in this file?

simon-smart88 (changes requested):
Very nearly there now - just this below and the `save_and_load` comment. I can't resolve the one on `cinema_schema.json` for some reason. Sorry if my reviewing has been a bit chaotic!

inst/shiny/modules/rep_cinema.R
```diff
@@ -0,0 +1,77 @@
...
+          configured_data = common[[input$data]],
+          gemtc_results = common[[gemtc_results()]]$mtcResults
+        ),
+        file
+      )
+    })
+
+  return(list(
+    save = function() {
+      shinyscholar::save_and_load(".", "rep_cinema")
```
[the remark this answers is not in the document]
tommorris168: Ah, I get it now!

simon-smart88 (approved):
Looks good!

I'm going to hold off merging as we need to update the deployment for the workshop next week first.

simon-smart88:
@JanionNevill @tommorris168 Could you check https://github.com/CRSU-Apps/MetaInsight/pull/254/commits/17fc605d2487a348b519adadd948658e05a64a36 ? That fixes the tests and I presumed that the previous was incorrect given it was duplicating the CI as PI, but I may not understand something.

tommorris168:
It was correct. In the fixed effect model the CI and the PI are equal. You wouldn't normally calculate a PI in this case, so metafor sets the PI to NA. But there was a reason why we did it, I think so we didn't have to split into fixed and random later on.

simon-smart88:
Hmm how come we have this then:

> imported_project$project$CM$contributionMatrices$hatmatrix$NMAresults[[1]]$`lower CI`
 [1]  0.0012 -0.0237 -0.0125 -0.0231 -0.0108 -0.0343 -0.0232 -0.0343 -0.0209
[10]  0.0073 -0.0040  0.0081 -0.0151 -0.0026  0.0073
> imported_project$project$CM$contributionMatrices$hatmatrix$NMAresults[[1]]$`lower PrI`
 [1] -0.0127 -0.0386 -0.0280 -0.0403 -0.0264 -0.0482 -0.0376 -0.0503 -0.0357
[10] -0.0078 -0.0208 -0.0072 -0.0314 -0.0177 -0.0057 

That's read directly from the `.cnm`

tommorris168:
I'm not sure what this file is. Is it output from CINeMA?

simon-smart88:
Yes (at least I think so). See https://github.com/CRSU-Apps/MetaInsight/blob/17fc605d2487a348b519adadd948658e05a64a36/tests/testthat/test-rep_cinema.R#L12

tommorris168:
Interesting. I've just gone through CINeMA and selected a fixed effect model, and it did indeed produce CIs and PIs that weren't the same. That should only happen in the random effects model. This warrants an e-mail to Theo and Adriani.

simon-smart88:
Thanks for investigating - are you okay to send it?

tommorris168:
I've found out what's going on. CINeMA creates the CI from the selected model (fixed or random), but the PI always comes from the random effects model. I guess it's because the CINeMA method requires an assessment of heterogeneity, which isn't possible from the FE model. Thanks for bringing this up, it may have gone unnoticed otherwise. I think your change is right ("right" in the sense of matching CINeMA, I'm not sure about "right" in the scientific sense!). I'll check the code as well, for both frequentist and Bayesian models.

simon-smart88:
Okay pending confirmation on PI, I think we are good to go now. I added checks to return errors if the dataset doesn't contain `rob` and `indirectness` - 99 % certain this is necessary as the data was not valid when those columns are missing and that's the whole point of cinema. I also added a note to guidance about that.

tommorris168:
I'm going to speak to Alex tomorrow about this. I'll get back to you afterwards.

simon-smart88:
Okay great. Added some more checks to stop it processing regression models and added to the vignette too. I also updated the link to cinema as the `.ch` one is redirecting to `.gr` now.

tommorris168:
There are three changes to prediction intervals in the last two commits:

- In the frequentist fixed effect model, there is now a PI which is centred on the FE point estimate but has length given by the RE model.
- In the Bayesian FE model, same as above, noting that the length of the PI is again determined by the frequentist (not Bayesian) RE model.
- In both Bayesian models, if the PI is smaller than the CI, then the PI is set to be equal to the CI.

simon-smart88 (commented):
This all looks good to me. Because CINeMA hasn't been updated though, the tests are failing because the json is different. Do we want to wait for them to update or accept that there is a slight difference for now? 

Something else I remembered we had discussed but not done yet is changing 'Reproduce' to 'Export'. We can do it in the UI very easily, but it would be better to rename the modules too. I can handle that, but conscious we should do it before CRAN and so unsure whether do to it on this branch or on another.

tommorris168:
You could make the UI change on a different branch so we don't have to wait for CINeMA.

simon-smart88:
Just noting here that `main` has a different version of `ReinstateTreatmentIds` with a `raw_label` parameter:

https://github.com/CRSU-Apps/MetaInsight/blob/1bcca8a4fdf358a3e99fc5aae942eeff2fa77dd7/R/data/data_wrangling.R#L339

It was used here:

https://github.com/CRSU-Apps/MetaInsight/blob/1bcca8a4fdf358a3e99fc5aae942eeff2fa77dd7/R/cinema/cinema_export.R#L72 

This version doesn't use the updated function though.

tpapak:
@simon-smart88, @tommorris168  — re the Should export NMA results test you re-enabled in 79158df: it's failing on the fixed-effect prediction interval, and I think the fix is to **drop the recentering** rather than work around the test.

  What's happening: the fixed-effect export shifts netmeta's prediction interval onto the FE point estimate:

  pi_adjustment <- nma_treatment_effect - freq_results$TE.random[row, col]
  lower_pi <- freq_results$lower.predict[row, col] + pi_adjustment
  upper_pi <- freq_results$upper.predict[row, col] + pi_adjustment

  CINeMA reads the raw lower.predict/upper.predict (no shift), so the exported PrI differs from CINeMA's reference by exactly TE.common − TE.random —
  that's the test failure.

  Why I'd just remove the recentering:

  - The pi_adjustment only exists in the fixed branch of .PrepareComparisonForCinema. MetaInsight's own results display (freq_summary_f.R) already writes the PI straight from net1$lower.predict/upper.predict with no shift — so the export is the only place doing this, and it's inconsistent with both MetaInsight's own UI and CINeMA.
  - A prediction interval is inherently a random-effects quantity anyway — netmeta returns a single τ²-based lower.predict/upper.predict, so the unadjusted value is the natural one to export.

  The change, in the fixed branch:

  lower_pi <- freq_results$lower.predict[row, col]
  upper_pi <- freq_results$upper.predict[row, col]

  This makes the test pass, aligns the export with MetaInsight's own display, and removes a one-off special case.

  One thing to confirm on your end: the gemtc fixed branch does the same shift plus a min/max clamp against the CI — if you drop the adjustment for the frequentist path, that one probably wants the same treatment for consistency.

tommorris168:
@tpapak @simon-smart88 
I thought we had agreed that you were going to do the recentring in CINeMA. If that's not the case, could you confirm with Adriani that she is happy to leave it as it is?

There are not currently any prediction intervals displayed in MetaInsight. We hope to add them to the forest plots in future, but it will only be for the random effects models. So there isn't any internal inconsistency within MetaInsight.

If we do use the random effects PI for the fixed effect model, I agree it makes sense to do it for both frequentist and Bayesian.

tpapak:
@tommorris168 @simon-smart88 
I confirmed with Adriani, and we suggest we do not do recentering in the export/import and just report what netmeta gives. I checked that this is what CINeMA also does currently so I find that is the cleanest way to got. Sorry for the confusion.

tommorris168:
@simon-smart88 @tpapak I've reverted the changes. The frequentist RE prediction interval is now used for both frequentist models and the Bayesian FE model.

simon-smart88 (approved):
Looks good. I'm not sure about when to merge it though with the CRAN submission in progress.

Issue #252, opened 2026-02-12
Interactive forest plot for excluding studies


simon-smart88:
In https://github.com/CRSU-Apps/MetaInsight/commit/0205c560226793c599c13d0b4bc9852e58929c6b I've put an initial prototype of a new interface for excluding studies from the sensitivity analysis. It reuses the `summary_study` plot but shows all the studies. When a line is clicked on, all the arms of that study are shaded red and an input value is updated containing the excluded studies. 

[...]

PR #260, opened 2026-03-03
Interactive study exclusion interface

simon-smart88:
See #252


inst/shiny/modules/setup_exclude.R
```diff
@@ -171,34 +150,55 @@ setup_exclude_module_server <- function(id, common, parent_session) {
...
-                                        choices = state$choices,
-                                        choicesOpt = list(disabled = state$disabled))
-      }
       updateRadioButtons(session, "effects", selected = state$effects)
     }
   ))
 })
 }
 
+setup_exclude_module_results <- function(id){
```
JanionNevill, line 186: This is somewhat outside the scope of this particular task, but I've been trying to work out what makes me uncomfortable about this accordion (I've had my reservations back when it was a table). I think I've worked out that it's the amount of space it takes up when open.

I'd like to hear your thoughts on changing it to be a modal instead. It'd still be accessible from everywhere, but it wouldn't dominate every view when left open. Not a change to make on this PR in any case.

tests/testthat/helper-other.R
```diff
@@ -0,0 +1,24 @@
...
+  app$set_inputs(tabs = "setup")
+  app$set_inputs(setupSel = "setup_reload")
+  app$upload_file("setup_reload-load_session" = path)
+  app$click("setup_reload-goLoad_session")
+  app$wait_for_js("$('.sweet-alert.visible').length > 0")
+  app$click(selector = ".confirm")
+}
+
+# click on the exclusions forest plot
+click_setup_exclude <- function(app, study){
```
JanionNevill, line 13: Just an FYI here: You can use `glue::glue()` here using `{{` for an escaped curly brace.

tests/testthat/test-setup_exclude.R
```diff
@@ -30,51 +31,106 @@ test_that("setup_exclude produces data of the correct type", {
...
+  # click on study lines
+  click_setup_exclude(app, "Leo")
+  click_setup_exclude(app, "Constantine")
+
+  app$wait_for_idle()
+  exclusions <- app$get_value(input = "setup_exclude-exclusions")
+  expect_setequal(exclusions, c("Leo", "Constantine"))
   app$wait_for_value(input = "setup_exclude-complete", ignore = list(NULL, "", "initial"))
 
+  # the JS updates the DOM rather than the output value, so this grabs as seen in the app
```
JanionNevill, line 107: I don't follow this comment. What is this part doing?

inst/shiny/www/js/exclusions.js
```diff
@@ -0,0 +1,81 @@
...
+
+        // Toggle study selection
+        var index = selectedStudies.indexOf(clickedStudy);
+        if (index > -1) {
+          selectedStudies.splice(index, 1);
+        } else {
+          selectedStudies.push(clickedStudy);
+        }
+
+        // Update opacity for all study lines
```
JanionNevill: Why does this update every line, not just the current line? This seems inefficient.

JanionNevill (changes requested):
A few comments and questions, but the functionality seems to work well 👍

inst/shiny/www/js/exclusions.js
simon-smart88: It's not every line, just those from the same study. I'll update the comment to make that clear.

tests/testthat/test-setup_exclude.R
simon-smart88: The JS isn't updating the value of `output$setup_exclude-plot` but the already rendered plot, so the usual `app$get_value(output = "setup_exclude-plot")` doesn't get the version after studies have been excluded. So this is needed to grab the version after it's manipulated by the JS.

inst/shiny/modules/setup_exclude.R
simon-smart88: Yes I know what you mean. I had suggested setting `max-height: 90vh` on all the plots so that they always fit on the screen which I think would help, but it would be nice to maintain a way that allows the user to see the updated results and the exclusion interface. I'll try to have a play around with some different layouts.

tests/testthat/helper-other.R
simon-smart88: Yes good point - updated.

inst/shiny/modules/setup_exclude.R
JanionNevill: Setting the maximum height would help a lot of cases, but would hamstring the app when trying to use large data sets. It'll be interesting to see if we get any user feedback about it.

JanionNevill (approved):
Looks good

PR #266, opened 2026-03-13
Reverse summary_study

tommorris168:


R/summary_study_f.R
```diff
@@ -499,12 +499,50 @@ summary_study_min_max <- function(pairwise, outcome){
...
 }
 
+
+#' Reverses the freq$d1 data frame created by meta::pairwise, as if the treatments had been the other way round.
+#' @param freq_d1 The $freq$d1 element of the common data.
+#' @inheritParams common_params
+#' @return Data frame
+#' @keywords internal
+#' @export
+ReverseFreqD1 <- function(freq_d1) {
```
tommorris168: I think I've gone overboard in this function, as I think the only things I needed to change were TE, treat1 and treat 2.

simon-smart88:
@tommorris168 I think that's everything. Having the separate function for calling `netmeta::netmeta` seemed a bit pointless so I merged that. There wasn't any need to export the treatment info when this is always available from `configured_data$treatments`. (I'm surprised that more tests didn't need refactoring)

R/summary_study_f.R
simon-smart88: Maybe, but seems sensible to reverse everything. 

Just to note about `@export` - that's only needed when the function is called in the app. For internal functions they should be tagged `@noRd` to stop them being added to the manual.

tommorris168:
@simon-smart88 I had intended that the reversing would be done in the parameters passed to meta::pairwise, not by my function, so I've updated it as such.

simon-smart88:
Aha yes that makes more sense!

PR #268, opened 2026-03-18
Duplicate studies in wide data

simon-smart88:
This is a fix for #263 - now if wide data has duplicate study names it gets flagged

### DIFF
--- a/R/setup_load_f.R
+++ b/R/setup_load_f.R
@@ -178,6 +178,11 @@ ValidateUploadedData <- function(data, outcome) {
     return(result)
   }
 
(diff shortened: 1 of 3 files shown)


tommorris168 (changes requested):
I didn't know about duplicated(). That would have helped in previous work.

Ironically, I believe the studies returned in the error message from .ValidateUniqueStudies can contain duplicates. You need to apply unique() somewhere for cases where the same study appears at least three times.

simon-smart88:
Yes good point. Fixed now.

tommorris168 (approved):

Issue #251, opened 2026-02-11
Use long data throughout


simon-smart88:
This seems to be the only place wide data is required given there is only one use of `LongToWide()`:

https://github.com/CRSU-Apps/MetaInsight/blob/1c9cab764fe7fca37df60d2d160a118821735643/R/freq_analysis.R#L14

So we _should_ be okay to use `WideToLong()` once in `setup_load()` and always store the data in long form. Then all the other `WideToLong()` can be removed. Then we can also remove this:

https://github.com/CRSU-Apps/MetaInsight/blob/db2d4996a532180583f84ab8b153afbaf7d625ca/R/setup_load_f.R#L737

which deals with removing {reshape2} in #245

JanionNevill:
`WrangleUploadedData()` doesn't seem to use `CleanTreatmentIds()`. Make sure that the function is called somewhere and it should be fine. I assume when you say "removed", you mean removed from the common object?

simon-smart88:
Sorry I wasn't being very precise - my point was that after this change the data should leave `setup_configure` in the same state whether it is uploaded as long or wide and so it made sense to refactor the tests to reuse the assumptions for long data when the wide data is uploaded. I'll open a PR and hopefully that will make it clearer to understand.

PR #269, opened 2026-03-18
Always use long data

simon-smart88:
As I noted in #251 this will currently break the shinyscholar branch, but won't once #260 is merged. Because of that the tests won't run at the moment, but once that is merged I will merge back into this and check everything runs smoothly.


tommorris168 (approved):
Seems fine. Hopefully it will merge OK as I see several of the more recent updates aren't here.

PR #271, opened 2026-03-25
Use subsetted data in meta-regression

simon-smart88:
This was discussed as something that would be nice to have. It was more complicated to implement than I first hoped, but I think this is working as intended now. The models are only responsive to the exclusions changing when the subsetted data is being used. If this looks okay, I will also add some tests to confirm everything behaves as expected.


tommorris168 (approved):
It all appears to work.
I've updated the Gelman statistic text in baseline risk.

PR #274, opened 2026-04-29
Regression parameters

tommorris168:
I've mixed up two of the commits. The one labelled "description" is actually adding to the tables and vice versa.


simon-smart88 (approved):
I can't comment on the stats, but after updating a couple of tests everything looks fine.

PR #286, opened 2026-05-14
Gelman preplot

simon-smart88:
Hopefully this sorts it.


RyanJField (changes requested):
Please remove the note from cran-comments.md

PR #287, opened 2026-05-14
Dev

RyanJField:
7.0.1 cran release


RyanJField:
This update removes the note causing Cran to reject the package by importing the gelman functions from Coda under GPL Licence and bumps the version to v7.0.1.

tommorris168 (approved):
Approved

Issue #290, opened 2026-06-04
`make_netconnect` reports incorrect number of studies


simon-smart88:
I just noticed that `make_netconnect` isn't reporting the correct number of studies. It seems related to us using `data = NULL` here:

https://github.com/CRSU-Apps/MetaInsight/blob/aa99a85774086554bc1fd9e852562c77a063dbde/R/summary_network_f.R#L91

We do that because R CMD check complained about the undefined variables. Turns out we can actually simplify it to just pass the data.

```
data <- setup_load("diabetes.csv", "binary") |>
  setup_configure("Placebo", "random", "OR", "good", 123)

connect <- make_netconnect(data$freq)
connect[["k"]]
length(unique(data$connected_data$Study))

connect2 <- netmeta::netconnection(treat1, treat2, studlab, data = data$freq$pairwise)
connect2$k

connect3 <- netmeta::netconnection(data$freq$pairwise)
connect3$k
```

PR #291, opened 2026-06-04
Fix netconnect

simon-smart88:
See #290 for details


tommorris168 (approved):
I couldn't see where this was going wrong before the change so I'll take your word for it.

PR #294, opened 2026-06-17
Switch to metaregression dropdown menu

simon-smart88:
This came up as a suggestion in the meeting last week. I'm ambivalent about it, but this is how it can be achieved if you want to do it.


RyanJField:
@simon-smart88 thanks, I'll demo it at the next meeting and go from there.

